PR #1, opened 2026-06-03
test: comprehensive test coverage, feat: tier-based model routing, fix: safety hardening

Rybens92:
## Summary

This PR adds tier-based model routing for workflow subagents, multiple safety
fixes around abort/persistence/resume paths, and comprehensive test coverage —
growing from 11 test files to **26 test files** with **610 passing tests**.

### Tier-Based Model Routing

- New `src/model-tier-config.ts` — workflow subagents use `opts.tier`
  (small/medium/big) resolved through a config file at `~/.pi/workflows/model-tiers.json`.
  Each tier stores exactly one model spec. Fresh install defaults all three tiers
  to the user's current Pi model — zero setup needed.
- New `src/workflows-models-command.ts` — interactive `/workflows-models` TUI
  using Pi's native SelectList component with scrollable model picker per tier.
- `WorkflowAgent.run()` resolves `tier` → config → `mainModel` → session default;
  `opts.model` is still supported and takes precedence.
- `opts.tier` surfaced in `modelRoutingGuideline()` in `workflow-tool.ts`.

### Safety Fixes

- `persistRun()` wrapped in try/catch with `console.warn` — disk-full no longer
  crashes the running workflow.
- Abort signal improvements in `WorkflowManager.executeRun()` — intentional
  stop preserves status set by `pause()/stop()`; `resume()` rejects aborted runs.
- `deliverText()` stale-context guard — unchanged reference between `/reload`
  cycles no longer blocks background result delivery.
- `installResultDelivery()` refreshes its Pi API reference on `/reload` instead
  of silently holding a stale closure.
- `.catch(() => {})` on background promise prevents Node unhandled-rejection on
  abort/pause/stop.

### Testing (610 tests, 0 failures)

- **15 new test files**: `agent.test.ts`, `agent-tier-resolution.test.ts`,
  `auto-workflow.test.ts`, `builtin-commands.test.ts`, `model-routing.test.ts`,
  `model-tier-config.test.ts`, `run-persistence.test.ts`,
  `structured-output.test.ts`, `utils.test.ts`, `web-tools.test.ts`,
  `workflow-display.test.ts`, `workflow-manager-abort.test.ts`,
  `workflow-saved.test.ts`, `workflow-tools-available.test.ts`,
  `workflows-models-command.test.ts`.
- **11 existing test files expanded**: `workflow-manager.test.ts` (1,462 lines),
  `workflow-ui.test.ts`, `workflow-editor.test.ts`, `task-panel.test.ts`,
  `saved-commands.test.ts`, `workflow-tool.test.ts`, `workflow-runtime.test.ts`,
  `workflow-parser.test.ts`, `workflow-commands.test.ts`,
  `builtin-workflows.test.ts`, `worktree.test.ts`.
- **Coverage includes**: model-tier config round-trip, tier resolution with
  fallbacks, all 17 parser validation errors, save/load/delete persistence with
  cold-start deserialization, abort/pause/stop/resume state machine (including
  failed-state transitions), worktree create/remove with real git, `parallel()`
  result ordering and empty input, concurrent background runs.
- **Zero `as any` casts** in test files — all type-safe.

### Other

- Saved workflows listed in `/workflows` navigator alongside active runs,
  with separate keybindings (delete saved, close navigator).
- `deliverText()` prioritizes `verdict` → `report` → `summary` → JSON fallback
  for cleaner background-delivery messages. Duration and token count shown.
- `workflow-tool.ts` updated guidelines and tool description.
- 44 files changed (+9,311 / −478).

### Backward Compatibility

All changes are additive. Existing workflow scripts and configurations continue
to work unchanged. No public API breaking changes.

---

*Sorry the PR is this big — it's all the changes I've been making for my own
setup. Most of it was written using the extension's own workflows.*


QuintinShaw:
Thanks for this, and sorry for the slow turnaround. There's a lot of genuinely useful stuff here, so rather than send it back for another round I went through it and pushed the cleanup straight to this branch. Wanted to write up what I changed and why.

The main thing: the tier routing wasn't actually working. `agent(prompt, { tier: 'small' })` was dropping the tier on the floor in the real run path. `AgentOptions` didn't have a `tier` field, and `runWorkflow` only ever passed `model:` down to `WorkflowAgent.run()`, never the tier, so `resolveAgentModelSpec` always got `tier: undefined` and everything fell back to the session default. The unit tests didn't catch it because they call the resolver directly, but a real workflow never hits it with a tier set. I only found it running an actual end-to-end workflow against a live model.

Fixed by adding `tier` to `AgentOptions`, passing it through to `run()`, and pinning down the precedence as explicit model > tier > phase model > default (if a tier is set with no explicit model, the phase model gets bypassed so the tier wins). The resume hash includes the tier now too. Tested it live: a configured `small` tier routed to the right model, `big` to another, and an explicit `model` still took priority. There's a fake-agent regression test for the wiring so this can't quietly break again.

While I was in there I cleaned up a few other things:

- Dropped the auto-classification code in `model-tier-config.ts` (the heuristics from before 3d4feec switched to single-model-per-tier — nothing calls them anymore), plus a dead `maxLogs` option and an unreachable markdown regex.
- The result-delivery `try/catch` in task-panel was wrapping a `void sendMessage()`, so it couldn't actually catch the rejection it was there for. Fixed that.
- `/workflows-models` was writing the config file just from opening the menu — now it only writes on "Save and exit".
- Deleting a saved workflow left its `/command` registered for the rest of the session (Pi has no unregisterCommand), so the handler now notices and tells you to reload instead of silently re-running it.
- Trimmed some redundant tests (a few assert(true) ones, the tier-config tests that were duplicated across three files, and a stack of near-identical constructor tests), shared a mock-pi helper, and switched the leftover `../dist/` imports to `../src/`. Also fixed a hardcoded absolute path and a couple of locale/timing assertions that only passed on the original machine.
- README + tool guideline now document `tier` and `/workflows-models`.

Everything's green (biome + tsc + the unit tests) and I've run the core paths and tier routing against a real model.

One thing worth flagging that isn't from this PR: structured output (`schema`) leans on the model actually calling the `structured_output` tool, and that was hit or miss with the models I tried — some didn't call it, one called it without filling the fields. The code does the right thing, it's more of a model-reliability / fallback question for later.

Going to squash-merge this with you as co-author and cut a 1.10.0 since it's a new feature. Thanks again, especially for writing so much of it with the extension's own workflows.

QuintinShaw:
Merged and released in v1.10.0 — now live on npm (`@quintinshaw/pi-dynamic-workflows@1.10.0`). Thanks again!

Rybens92:
Thanks for taking a look and merging!
I know it wasn't easy to review PR this big, so I appreciate it very much!

PR #3, opened 2026-06-08
fix: truncate task panel lines to terminal width

xuli500177:
### Problem
When a workflow has a long name or phase name, the task panel renders a line exceeding the terminal width, causing pi-tui to crash with:

```
Error: Rendered line 440 exceeds terminal width (69 > 45).
```

### Root Cause
`renderPanel()` builds status lines without respecting the terminal width:

```typescript
return `  ${icon} ${r.workflowName}  ${done}/${agents.length} agents${phase}`;
```

When `workflowName` + `phase` is long, the line overflows.

### Fix
1. Pass `width` from the `Component.render()` callback to `renderPanel()`
2. Truncate lines that exceed `maxWidth`

The lines in `rows` are intentionally plain text (no ANSI codes), so `string.length` is sufficient for width calculation.

### DIFF
--- a/src/task-panel.ts
+++ b/src/task-panel.ts
@@ -111,6 +111,7 @@ export function renderPanel(manager: WorkflowManager, theme: Theme, width?: numb
     const icon = r.status === "paused" ? "⏸" : "◆";
     const phase = live?.snapshot.currentPhase ? ` · ${live.snapshot.currentPhase}` : "";
     const line = `  ${icon} ${r.workflowName}  ${done}/${agents.length} agents${phase}`;
+    // Truncate line to fit terminal width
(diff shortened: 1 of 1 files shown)


QuintinShaw:
Thanks for the fix direction here. I agreed with the root cause, but took a slightly stricter implementation in #8: it uses Pi TUI's `visibleWidth` / `truncateToWidth`, covers ANSI and wide-character cases, and adds a 42-column regression test plus a real Pi smoke.

I'll close this as superseded once #8 is merged.

QuintinShaw:
Closing as superseded by #8, which has now been merged. Thanks again for identifying the root cause and sending the initial fix direction.

PR #13, opened 2026-06-09
feat(ui): improve workflow display theming and wrapping

matheuswidder:
## Summary

Theme-aware rendering for the workflow display, ANSI-safe text wrapping for the navigator overlay, and visual box borders with focus state.

### Changes

1. **Theme-aware `renderWorkflowLines()`** — accepts an optional `ThemeLike` parameter so the workflow display renders with the active Pi theme (accent colors, bold, dim) instead of plain text. Works in both the TUI widget (real Theme) and tool output / tests (identity passthrough).

2. **ANSI-safe wrapping in the navigator** — replaced the manual character-slice `wrap()` with `wrapTextWithAnsi()` from `@earendil-works/pi-tui`, so wrapped text respects ANSI escape sequences.

3. **Box border overlay** — the workflow navigator now renders inside a `╭─╮││╰─╯` border box that becomes brighter when focused, following the same pattern as `pi-ask-user`.

4. **Sidebar anchor support** — `NavigatorOptions.anchor` lets the overlay display as a right-side sidebar (`anchor="right-center"` at 60% width) instead of centered.

5. **Focusable component** — the navigator component now implements the `Focusable` interface, tracking focused state for border accent.

### Render invalidation fix

`createWidgetWorkflowDisplay()` now re-registers the widget factory on every `update()` and `complete()` call, ensuring the TUI re-renders the workflow widget even when `showStatus=false` (the default).

### Validation

- `npm test`: ✓ (biome check + tsc + 651/651 unit tests)
- Regression test added: "re-renders via setWidget even when showStatus is false (default)"
- `.context/` removed from git tracking (auto-generated files unrelated to PR scope)
- `.gitattributes` added (`* text=auto`) for line ending normalization

### Files changed

```
 .gitattributes                     |   1 +
 .gitignore                         |   1 +
 src/display.ts                     |  23 +-
 src/workflow-ui.ts                 |  27 +-
 tests/workflow-display.test.ts     |  55 ++-
 tests/workflow-saved.test.ts       |   3 +-
 6 files changed, 73 insertions(+), 2494 deletions(-)
 40 files deleted (.context/)
```


QuintinShaw (changes requested):
Thanks for the contribution. I reviewed this against the current `main` and ran the repository validation locally.

This PR is not ready to merge yet:

1. `npm test` currently fails at `biome check .`.
   - `npm run build` passes.
   - `npm run test:unit` passes, 650/650.
   - But the full repository validation must pass before merge.

2. The PR includes a large generated `.context/*` tree that appears unrelated to the UI/theme/ANSI wrapping changes. Please remove those files from this PR so the scope stays focused.

3. The `createWidgetWorkflowDisplay()` change looks risky. It registers the widget component once, then later `update()` / `complete()` only mutate closed-over state. With the default `showStatus=false`, I do not see a render invalidation path, so the workflow widget may stop updating live. Please either preserve the previous update behavior or add a clear render invalidation mechanism with a regression test.

4. The PR description also includes already-merged work from #10, #11, and #12. Please rewrite it to describe only the actual changes in this PR.

Once these are addressed, I’m happy to review again.

matheuswidder:
All four issues from the review have been addressed:

1. ✅ **`npm test` now passes** — biome check (0 errors), tsc, and 651/651 unit tests all green.

2. ✅ **`.context/` removed** — the auto-generated AICoders context tree has been deleted from git tracking and added to `.gitignore`. A `.gitattributes` file (`* text=auto`) was also added for line ending normalization.

3. ✅ **Render invalidation in `createWidgetWorkflowDisplay()`** — the widget factory is now re-registered on every `update()` and `complete()` call via `ctx.ui.setWidget(key, widgetFactory, {placement})`, ensuring the TUI re-renders the workflow widget even when `showStatus=false`. A regression test with a render spy was added to confirm the invalidation works.

4. ✅ **PR description rewritten** — only describes the actual changes (theme-aware display, ANSI-safe wrapping, border box, sidebar anchor, render invalidation fix).

---

Commits since the review:
- `c82da8b` — fix: address PR review (biome check, remove .context/, render invalidation, regression test)
- `789861e` — test: strengthen render invalidation test with render spy

Ready for re-review. Thanks!

QuintinShaw (approved):
Thanks for the quick follow-up and for cleaning this up.

Good job narrowing the PR scope and adding the render invalidation regression test. I re-reviewed the latest revision and verified it locally with `npm test`: biome check, tsc, and 651/651 unit tests all pass.

The scoped diff looks good now: the generated `.context/` files are gone, the widget render invalidation issue is covered, and the UI changes are focused on theming/wrapping.

I adjusted the PR title to match the final scope and will merge this with a squash commit.

PR #29, opened 2026-06-22
Make workflow triggering configurable

Whamp:
I kept accidentally triggering workflows for trivial things because I use that word a lot in my normal conversations with an agent. I didn't see an existing way to turn off the `workflow` keyword trigger and still manually trigger a workflow. So this PR ships two changes to improve this behavior. Feel free to take or leave any of it, no hard feelings.

Change 1 lets the trigger word be user configurable. For me, I have changed it to `pi-workflow` in my branch and that works infinitely better to avoid accidents.

Change 2, which is almost unnecessary now that the user trigger word is user-configurable, is the creation of a manual trigger for a workflow in case people want the automatic phrase-based trigger disabled but still want to use a workflow:

```text
/workflows run <prompt-describing-workflow-to-run>
```

## Summary

- Adds `keywordTriggerWord` setting with `/workflows-trigger set <word>` and `/workflows-trigger reset`.
- Preserves default `workflow` / `workflows` behavior for existing users.
- Makes custom trigger words literal, case-insensitive terms and excludes slash-command forms.
- Adds `/workflows run <prompt>` as an explicit manual trigger that works even when keyword triggering is off.
- Updates README documentation and regression tests.

## Validation

- `npm test` — 740 tests passing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,11 @@ Ask in plain language:
 Run a workflow to audit every route under src/routes/ for missing auth checks.
 ```
 
-Pi writes the script and runs it in the background — your turn ends immediately and a live panel tracks progress while you keep working. Or just type the word **workflows** in any message to force one. If you only want to discuss workflows without triggering one, run `/workflows-trigger off`; the preference is saved for new sessions in `~/.pi/workflows/settings.json`. Check the current state with `/workflows-trigger status`, and turn it back on with `/workflows-trigger on`.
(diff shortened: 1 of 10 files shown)


QuintinShaw:
Thanks @Whamp, this is a clean fix and it solves exactly the accidental-trigger pain a few of us kept running into. Verified locally — build's green, 740/740 tests pass, and I smoke-tested `/workflows run` end-to-end against a real pi session with the keyword trigger turned off; it forces the workflow turn correctly. Merging now

Whamp:
Awesome thanks for the review and consideration.

PR #30, opened 2026-06-22
Claude-Code-parity /workflows TUI + workflow prompting improvements

paulbrav:
Two related sets of changes that bring pi-dynamic-workflows closer to Claude Code's dynamic-workflows experience.

## `/workflows` TUI — two-pane layout

Restyle the navigator's **phases** and **agents** views to match Claude Code's `/workflows`:

- a left **Phases** box and a right **`<phase> · N agent`** box sharing one top/bottom border (`┌ … ┬ … ┐` / `└ … ┴ … ┘`)
- a `›` selection caret and `● name  model  Nk tok` agent rows with right-aligned token counts
- a two-line header: run name, then status + right-aligned `done/total agents · tokens`

Width-safe — collapses to a single full-width pane below ~37 columns. The **runs** and **detail** views and all key bindings are unchanged; the one render snapshot test is updated to the new layout.

After:
```
audit
running                                                     2/3 agents · 150 tok
┌ Phases ──────┬─ Scan · 2 agents ─────────────────────────────────────────────┐
│  1 Scan 2/2  │› ● scan a  model                                       100 tok│
│  2 Report 0/1│  ● scan b  model                                        50 tok│
└──────────────┴───────────────────────────────────────────────────────────────┘
```

## Prompting

- **agent.ts** — give *every* subagent (not only schema agents) the framing that it is a subagent whose only returned result matters and that it has no parent/repo context it wasn't given.
- **workflow.ts** — `verify()` defaults move from 2 reviewers / 0.5 to **3 / 0.66**, so a 1-1 tie no longer passes (a clear majority is required to confirm a finding).
- **workflow-tool.ts** — document the default **finder → verify → merge/rank/cap** fan-out shape, and that each subagent should get a substantive task rather than a trivial one-file read.
- **effort-command.ts** — note ultracode is session-only / not persisted, and that exhaustive runs should set explicit token/agent caps.

## Verification

- `npm run build` (tsc): clean
- `npm run test:unit`: 725/725 pass
- `biome check`: clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


QuintinShaw (changes requested):
Thanks for this, @paulbrav — the two-pane navigator looks great and the prompting tweaks are sensible. Build and unit tests are green as-is; a few things I'd like to see before merging:

1. Single-pane phase coloring (`src/workflow-ui.ts`). `renderSinglePane` calls `leftPhaseRow(..., [], ...)` with an empty agents array, so `phaseStatusColor` can't see running/error state and those phases fall back to dim. The two-pane path passes `model.agents(runId, p.title)` — could you thread the per-phase agents through here the same way? Completed phases still color correctly (that uses `p.done/total`), so it's only running/error that's affected.

2. Width-safety test. The new phases/agents render tests all run at width 80, i.e. the two-pane path. Since the headline feature is the sub-37-column collapse, could you add a test that renders below the single-pane threshold and asserts the fallback to `renderSinglePane` (full-width, no two-pane divider)? That path is currently uncovered.

3. verify() default change. Bumping reviewers 2 to 3 / threshold 0.5 to 0.66 is a good call — a 1-1 tie shouldn't confirm. Since it changes behavior and adds ~50% reviewer cost per verify for default callers, could you just call it out in a CHANGELOG/release-note line so it isn't a silent default change?

None of these are blockers on the functionality — they're about covering the new render path and not shipping a silent default change. Happy to merge once they're in.

QuintinShaw:
The TUI work here is high quality — I built the merge commit against current main and got tsc clean, 746/746 green, and the width behavior (37-col collapse, shared frame) checks out exactly as described. Also glad to see the header sums per-agent tokens rather than run-level `tokenUsage`, which stays 0 until run-end.

Two asks:

1. The `verify()` default change (2 reviewers/0.5 → 3/0.66) is a silent +50% verify cost for every existing script that calls `verify()` without opts, and it has a sharp edge: if one of the 3 reviewers errors and gets filtered out, 1/2 = 0.5 < 0.66, so a partial failure effectively demands a unanimous 2/2. Could you split that change (plus the unconditional subagent framing) into its own PR so the TUI can land now? Happy to discuss the defaults there — the 1-1 tie fix itself is a good catch.
2. Please update the branch — it's behind main, and #45 (which also touches `workflow-ui.ts`) will be rebased on top of this once it lands.

Minor notes, non-blocking: persisted runs render "0 tok" per agent (per-agent tokens aren't persisted — pre-existing gap your layout makes visible), and the scroll indicator only shows overflow below, not above.

QuintinShaw:
Landing this now. I pushed the integration directly to your branch (`maintainerCanModify`): merged current `main` to clear the conflicts from #43/#49, and split out the two things from the review so the TUI ships clean — reverted the `verify()` default change (back to 2 reviewers / 0.5) and the unconditional subagent framing in `agent.ts`. Kept everything else, including the effort-command and workflow-tool prompting tweaks.

The two-pane renderer is really nice work — 757/757 green, tsc clean, and I ran a real end-to-end workflow against it to confirm the wiring. The `verify()` 1-1-tie observation was a good catch; feel free to open a follow-up PR for the defaults and we can settle on numbers there. Thanks for the contribution!

Issue #26, opened 2026-06-20
Fails to handle running out of tokens


SevereOverfl0w:
If you run out of your subscription budget, then the workflow fails to "checkpoint" and be resumable:

```                                                                                                                                                                                                                  
 [workflow-result]
 ✗ Background workflow mql6y23z-gqkbfc failed: Subagent did not produce valid structured_output after repair attempts                                                                                             

 Error: Codex usage limit reached (plus plan). Resets in ~3h. Current premium: 5h 100%, weekly 42%.
```

and then upon tokens refilling the agent has to start completely afresh.

QuintinShaw:
Thanks for the great report — and for including the exact output. That kind of detail is what made this easy to pin down.

You read the symptom exactly right: hitting a usage limit shouldn't be a terminal failure, and it certainly shouldn't discard work that's already done. What was happening under the hood: when a usage/quota limit is reached mid-run, the provider doesn't raise an error — it quietly ends the turn with an error status on the message. The workflow runner wasn't inspecting that, so it misread the empty result as a schema-compliance problem (`did not produce valid structured_output`) and marked the whole run failed. The journal of completed steps was actually being saved the entire time; it just wasn't surfaced, so there was no signal that the run could be picked back up.

The fix (#28) teaches the runner to recognize a provider usage/quota/rate limit for what it is — a transient, recoverable condition. Instead of failing, the run now checkpoints as `paused`, keeps its journal, and tells you how to continue:

```
⏸ Background workflow <id> paused: usage limit reached (resets in ~3h).
Completed steps are saved — run /workflows resume <id> once your limit resets.
```

Resuming replays everything that already finished and only re-runs the remainder, so nothing is lost. It's covered by tests, including one that drives a real subagent session into the limit so this path stays guarded.

One thing I deliberately left out of this pass: automatically resuming the moment your quota refills. That needs a little more care — scheduling, restart-safe timers, and a back-off so it doesn't just bounce off the limit again — so I've split it into a follow-up (#27). The reset time is already persisted, so it's ready to build on; for now, resume is a single command once you're back under your limit.

Appreciate you flagging this!

PR #43, opened 2026-06-27
feat: wire agent frontmatter isolation to worktree execution

gr3enarr0w:
## Summary

Implements upstream #37 by wiring `isolation: worktree` from agent frontmatter into workflow agent execution.

- parses `isolation: worktree` in agent definitions
- includes isolation in `agentDefinitionKey` so resume cache invalidates when isolation changes
- resolves isolation from explicit agent options first, then agent definition frontmatter
- annotates runner instructions with the resolved isolation source
- adds runtime coverage proving agent-type isolation uses an isolated worktree cwd

Closes #37.

## Review notes

Full scoped review completed before publishing. No additional findings remained after adding runtime coverage for frontmatter-driven isolation.

## Verification

- `npx biome check src/agent-registry.ts src/workflow.ts tests/agent-registry.test.ts`
- `npm run build`
- `npm run test:unit -- tests/agent-registry.test.ts tests/worktree.test.ts` (rebased branch full project script: 746 passing, 0 failing)

PR #42, opened 2026-06-27
feat: add built-in /code-review workflow command

gr3enarr0w:
## Summary

Implements upstream #32 by adding a built-in `/code-review` workflow command.

- adds `generateCodeReviewWorkflow()` with 7 parallel review angles, verification, and final report phases
- registers `/code-review` alongside the existing built-in workflow commands
- defaults an empty command argument to the current directory
- handles missing/empty gather output without throwing
- adds parser and command registration coverage

Closes #32.

## Review notes

Full scoped review completed before publishing. One issue was found and fixed before PR creation: the gather phase now handles `null`/missing output instead of dereferencing `gathered.diff`.

## Verification

- `npx biome check src/code-review.ts src/builtin-commands.ts tests/builtin-commands.test.ts tests/builtin-workflows.test.ts`
- `npm run build`
- `npm run test:unit -- tests/builtin-workflows.test.ts` (project script ran full unit suite: 745 passing, 0 failing)

PR #46, opened 2026-06-27
feat: auto-resume on provider usage limit reset + /code-review command

### LINKED ISSUE #27: Auto-resume a workflow once a provider usage limit resets
Follow-up to #26.

#26 makes a run that hits a provider usage/quota limit **checkpoint as `paused`** (instead of failing), preserving its journal so `/workflows resume <id>` replays the completed prefix and runs only the remainder. Resuming is currently **manual**.

This issue tracks **automatic** resume once the quota refills.

### What's needed
- A scheduler / restart-safe timer that fires when the limit's reset window elapses.
- A back-off + attempt cap so a run that resumes into a still-exhausted budget doesn't bounce off the wall repeatedly (it would just re-pause).
- Re-arming the timer across process restarts (a paused-for-usage-limit run should reschedule on cold start).
- Decide whether auto-resume is default-on (with an opt-out) or opt-in via `ExecOptions`.

### Hook already in place
The fix persists the pause reason and the provider's reset hint on the run state (`pauseReason: "usage_limit"`, `resetHint`), and `recoverStaleRuns()` already reconciles stale runs on startup — so the scheduling layer has what it needs to build on.

### Out of scope here
Parsing every provider's reset-time phrasing into an absolute timestamp — the current hint is kept verbatim; a scheduler can parse best-effort and fall back to a conservative default delay.

### LINKED ISSUE #32: feat: add /code-review built-in workflow command
## Summary

Add a `/code-review` built-in slash command that runs a structured multi-angle code review using parallel workflow agents. This is the most-requested missing capability — the package has `/adversarial-review` (generic investigate→refute) and `/deep-research` (web research) but no code-specific review workflow.

## Problem

`/adversarial-review` is generic: one agent investigates, N agents try to refute each finding. For code review this misses the most valuable pattern: **specialized per-angle finders** that each hunt for a different bug class, then a verify pass per candidate. A single investigate agent scanning a diff for all bug types simultaneously produces shallower findings than 7 agents each deeply specialized in one angle.

### New file: `src/code-review.ts`

```ts
export function generateCodeReviewWorkflow(): string
```

Returns a workflow script string (same pattern as `generateAdversarialReviewWorkflow`) implementing:

**Phase 1 — Find (7 parallel agents, one per angle)**

| Angle | Tier | Focus |
|---|---|---|
| A — line-by-line scan | medium | Inverted conditions, off-by-one, nil deref, wrong variable, swallowed error |
| B — removed-behavior audit | medium | Every deleted line: name the invariant, find where re-established |
| C — cross-file tracer | medium | Changed functions → grep callers → check call site breakage |
| D — reuse | small | New code duplicating existing helpers |
| E — simplification | small | Redundant state, copy-paste variation, dead code |
| F — efficiency | small | Redundant I/O, sequential work that could parallel, startup blocking |
| G — altitude | big | Fix at wrong abstraction level; bandaid on shared infrastructure |

Each finder returns `{ candidates: Array<{ file, line, summary, failure_scenario }> }` via structured output.

**Phase 2 — Verify (parallel, one agent per unique candidate)**

Uses `verify()` stdlib with `reviewers: 1, threshold: 0.5`. Each verifier gets the diff + relevant file(s) + candidate and returns `CONFIRMED / PLAUSIBLE / REFUTED`. Drop REFUTED.

**Phase 3 — Report (big tier)**

Synthesis agent ranks surviving candidates by severity, caps at 10, produces:
- Ranked JSON array: `[{ file, line, summary, failure_scenario }]`
- Human-readable markdown report

### Command registration in `src/builtin-commands.ts`

```ts
// alongside /deep-research and /adversarial-review
pi.registerCommand('code-review', {
  description: 'Multi-angle parallel code review: 7 finder angles + verify pass → ranked findings',
  usage: '/code-review [diff-range|file-path|PR-number]',
  handler: async (args, ctx) => { ... }
})
```

### Input modes

- `/code-review` — reviews `git diff HEAD` (working tree + staged)
- `/code-review HEAD~3..HEAD` — reviews a git range
- `/code-review src/foo.ts` — reviews a specific file
- `/code-review 42` — if a number, attempts `gh pr diff 42`

### Model tier routing

Finders A/B/C → `medium`, D/E/F → `small`, G → `big`, synthesis → `big`

## Implementation notes

- Finder agents each read the diff + relevant file context via the `read`/`grep` tools
- `structured_output` schema per finder: `{ candidates: [{ file: string, line: number, summary: string, failure_scenario: string }] }`
- Dedup before verify: same file+line+reason → keep one
- The verify pass uses `parallel(candidates.map(c => () => verify(c, { reviewers: 1 })))` 
- Correctness findings (A/B/C) always outrank cleanup (D/E/F) and altitude (G) when cap forces a cut

## Files to change

- `src/code-review.ts` — new file
- `src/builtin-commands.ts` — register `/code-review`
- `src/index.ts` — export `generateCodeReviewWorkflow`
- `README.md` — document the new command

gr3enarr0w:
## Summary

Closes #27
Closes #32

---

## feat(workflow-manager): auto-resume after usage-limit pause (#27)

When a workflow hits a provider quota/usage limit, it's now checkpointed as `paused` (already in place via #26). This PR adds the **automatic** resume layer on top.

### How it works

- `ExecOptions.autoResume` (default `true`) — opt out per-execution by passing `autoResume: false`
- When `executeRun` catches a `PROVIDER_USAGE_LIMIT` error, it calls `scheduleAutoResume(runId, resetHint)`
- `parseResetHintMs()` extracts hours/minutes from the verbatim `resetHint` string (e.g. `"~3h"`, `"45 min"`), falls back to **60 minutes**
- Exponential backoff: each re-pause doubles the delay; `MAX_AUTO_RESUME_ATTEMPTS = 5` then gives up
- `recoverStaleRuns()` now re-arms timers on cold start for any persisted `paused + pauseReason === "usage_limit"` runs
- `stop()` and `deleteRun()` cancel pending timers so stopped runs don't ghost-resume
- Emits `"autoResuming"` event before each attempt for observability

---

## feat(code-review): /code-review built-in command (#32)

New `src/code-review.ts` implementing `generateCodeReviewWorkflow()`, registered as `/code-review`.

### Phases

**Phase 1 — Find (7 parallel agents)**

| Agent | Tier | Focus |
|---|---|---|
| A — line-by-line scan | medium | Inverted conditions, off-by-one, nil deref, wrong variable, swallowed error |
| B — removed-behavior audit | medium | Every deleted line: invariant named, where re-established |
| C — cross-file tracer | medium | Changed functions → grep callers → call site breakage |
| D — reuse | small | New code duplicating existing helpers |
| E — simplification | small | Redundant state, copy-paste variation, dead code |
| F — efficiency | small | Redundant I/O, sequential work that could parallel |
| G — altitude | big | Fix at wrong abstraction level |

**Phase 2 — Verify** — parallel verifier per candidate, drops REFUTED

**Phase 3 — Report** — big-tier synthesis, ranked A/B/C > D/E/F > G, capped at 10

### Input modes

```
/code-review                  → git diff HEAD
/code-review HEAD~3..HEAD     → git range
/code-review src/foo.ts       → specific file
/code-review 42               → gh pr diff 42
```

### DIFF
--- a/src/builtin-commands.ts
+++ b/src/builtin-commands.ts
@@ -4,7 +4,7 @@
  * They run a generated workflow script and print the final report.
  */
 
-import { exec } from "node:child_process";
+import { execFile } from "node:child_process";
 import { promisify } from "node:util";
 import { createCodingTools, type ExtensionAPI, type ExtensionCommandContext } from "@earendil-works/pi-coding-agent";
 import { generateAdversarialReviewWorkflow, generateMultiPerspectiveWorkflow } from "./adversarial-review.js";
@@ -13,7 +13,7 @@ import { generateCodebaseAuditWorkflow, generateDeepResearchWorkflow } from "./d
 import { createWebTools } from "./web-tools.js";
 import { runWorkflow, type WorkflowRunResult } from "./workflow.js";
 
-const execAsync = promisify(exec);
+const execFileAsync = promisify(execFile);
 
 function alreadyRegistered(pi: ExtensionAPI, name: string): boolean {
   try {
@@ -131,16 +131,26 @@ export function registerBuiltinWorkflows(pi: ExtensionAPI, opts: { cwd: string }
         let diff = "";
 
         try {
+          let cmd: string;
+          let cmdArgs: string[];
           if (!input) {
             diffSource = "git diff HEAD";
+            cmd = "git";
+            cmdArgs = ["diff", "HEAD"];
           } else if (/^\d+$/.test(input)) {
             diffSource = `gh pr diff ${input}`;
+            cmd = "gh";
+            cmdArgs = ["pr", "diff", input];
           } else if (input.includes("..")) {
             diffSource = `git diff ${input}`;
+            cmd = "git";
+            cmdArgs = ["diff", input];
           } else {
             diffSource = `git diff HEAD -- ${input}`;
+            cmd = "git";
+            cmdArgs = ["diff", "HEAD", "--", input];
           }
-          const { stdout } = await execAsync(diffSource, { cwd });
+          const { stdout } = await execFileAsync(cmd, cmdArgs, { cwd });
           diff = stdout;
           if (!diff.trim()) {
             return ctx.ui.notify(`No diff output from: ${diffSource}`, "warning");
--- a/src/builtin-commands.ts
+++ b/src/builtin-commands.ts
@@ -4,11 +4,16 @@
  * They run a generated workflow script and print the final report.
  */
 
+import { exec } from "node:child_process";
+import { promisify } from "node:util";
 import { createCodingTools, type ExtensionAPI, type ExtensionCommandContext } from "@earendil-works/pi-coding-agent";
 import { generateAdversarialReviewWorkflow, generateMultiPerspectiveWorkflow } from "./adversarial-review.js";
+import { generateCodeReviewWorkflow } from "./code-review.js";
 import { generateCodebaseAuditWorkflow, generateDeepResearchWorkflow } from "./deep-research.js";
 import { createWebTools } from "./web-tools.js";
 import { runWorkflow, type WorkflowRunResult } from "./workflow.js";
+
+const execAsync = promisify(exec);
 
 function alreadyRegistered(pi: ExtensionAPI, name: string): boolean {
   try {
@@ -111,6 +116,55 @@ export function registerBuiltinWorkflows(pi: ExtensionAPI, opts: { cwd: string }
         } catch (error) {
           ctx.ui.setStatus("multi-perspective", undefined);
           ctx.ui.notify(`multi-perspective failed: ${error instanceof Error ? error.message : error}`, "error");
+        }
+      },
+    });
+  }
+
+  if (!alreadyRegistered(pi, "code-review")) {
+    pi.registerCommand("code-review", {
+      description:
+        "Multi-angle parallel code review: 7 specialized finders (correctness, reuse, simplification, efficiency, altitude) + verify pass → ranked findings",
+      async handler(args: string, ctx: ExtensionCommandContext) {
+        const input = args.trim();
+        let diffSource = "git diff HEAD";
+        let diff = "";
+
+        try {
+          if (!input) {
+            diffSource = "git diff HEAD";
+          } else if (/^\d+$/.test(input)) {
+            diffSource = `gh pr diff ${input}`;
+          } else if (input.includes("..")) {
+            diffSource = `git diff ${input}`;
+          } else {
+            diffSource = `git diff HEAD -- ${input}`;
+          }
+          const { stdout } = await execAsync(diffSource, { cwd });
+          diff = stdout;
+          if (!diff.trim()) {
+            return ctx.ui.notify(`No diff output from: ${diffSource}`, "warning");
+          }
+        } catch (err) {
+          return ctx.ui.notify(
+            `Failed to get diff (${diffSource}): ${err instanceof Error ? err.message : err}`,
+            "error",
+          );
+        }
+
+        ctx.ui.notify(`Reviewing diff (${diffSource}) — running 7 finder angles in parallel…`, "info");
+        try {
+          const result = await runWorkflow(generateCodeReviewWorkflow(), {
+            cwd,
+            args: { diff, diffSource },
+            tools: createCodingTools(cwd),
+            onPhase: (title) => ctx.ui.setStatus("code-review", `review: ${title}`),
+          });
+          ctx.ui.setStatus("code-review", undefined);
+          await pi.sendMessage({ customType: "code-review", content: reportText(result), display: true });
+        } catch (error) {
+          ctx.ui.setStatus("code-review", undefined);
+          ctx.ui.notify(`code-review failed: ${error instanceof Error ? error.message : error}`, "error");
         }
       },
     });
--- a/src/code-review.ts
+++ b/src/code-review.ts
@@ -1,0 +1,155 @@
+/**
+ * Multi-angle parallel code review workflow.
+ * 7 specialized finder agents → verify pass → ranked report.
+ */
+
+/**
+ * Generate a code-review workflow script.
+ *
+ * The workflow expects `args` to be passed with shape:
+ *   { diff: string, diffSource: string }
+ *
+ * Model tier routing follows the spec:
+ *   Finders A/B/C → medium (correctness)
+ *   Finders D/E/F → small  (cleanup)
+ *   Finder  G     → big    (altitude / abstraction)
+ *   Synthesis     → big
+ */
+export function generateCodeReviewWorkflow(): string {
+  return `export const meta = {
+  name: 'code_review',
+  description: 'Multi-angle parallel code review: 7 finder angles + verify pass → ranked findings',
+  phases: [
+    { title: 'Find' },
+    { title: 'Verify' },
+    { title: 'Report' },
+  ],
+}
+
+const diff = (args && args.diff) || ''
+const diffSource = (args && args.diffSource) || 'git diff HEAD'
+const candidateSchema = {
+  type: 'object',
+  properties: {
+    candidates: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          file: { type: 'string' },
+          line: { type: 'number' },
+          summary: { type: 'string' },
+          failure_scenario: { type: 'string' },
+        },
+        required: ['file', 'line', 'summary', 'failure_scenario'],
+      },
+    },
+  },
+  required: ['candidates'],
+}
+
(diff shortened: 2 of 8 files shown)


QuintinShaw:
Thanks — the auto-resume work is the valuable part here, but this PR is doing three separate things and I can't take it as one:

- agent-registry.ts / workflow.ts / agent-registry.test.ts are byte-identical to #43 (same isolation wiring) — I'm landing #43 for that.
- The /code-review command overlaps #42; see my note there. We're consolidating on this generator, but it needs tests — it currently breaks `commands.length === 4`.
- Auto-resume (#27) is the only unique change, and it ships with no tests.

Could you rebase this down to just auto-resume + /code-review, drop the isolation files (they'll come from #43), and add coverage for parseResetHintMs, the backoff/cap in scheduleAutoResume, and the recoverStaleRuns re-arm? Happy to merge once it's scoped to its own change.

QuintinShaw:
Thanks for the quick turnaround — the split is clean (no more #43 overlap, rebases onto main with zero conflicts) and the new tests are real. Two things before merge, both in the new code:

1. stop() can't cancel an auto-resume timer that recoverStaleRuns() arms on cold start. The `if (!managed) return false` guard returns before cancelAutoResume() runs, and recovered runs aren't in this.runs — so a run the user explicitly stops can still silently auto-resume and re-spend quota. deleteRun() already cancels unconditionally; stop() should mirror it (move the two cancel lines above the guard).

2. The /code-review diff-source runs through exec() (/bin/sh -c) with the arg interpolated raw in the `..` and filename branches: `git diff ${input}` / `git diff HEAD -- ${input}`. That's a shell-injection footgun. worktree.ts already uses promisify(execFile) + array args — please follow that here.

A regression test asserting both stop() and deleteRun() call cancelAutoResume would pin #1 (that gap is exactly why it slipped through). Optional: clamp the backoff with Math.max(delay, 60_000) so a hint that parses to ~0 doesn't fire five instant retries.

gr3enarr0w:
## PR Reviewer Guide 🔍

#### (Review updated until commit https://github.com/QuintinShaw/pi-dynamic-workflows/commit/2347f1af4ebdbd025453271833a748ab39ca8d04)


Here are some key observations to aid the review process:

<table>
<tr><td>

**🎫 Ticket compliance analysis 🔶**



**[27](https://github.com/QuintinShaw/pi-dynamic-workflows/issues/27) - Partially compliant**

Compliant requirements:

- ExecOptions.autoResume defaults to true with per-execution opt-out
- executeRun calls scheduleAutoResume on PROVIDER_USAGE_LIMIT
- parseResetHintMs extracts hours/minutes and falls back to 60 minutes
- Exponential backoff doubles the delay each re-pause
- MAX_AUTO_RESUME_ATTEMPTS = 5
- recoverStaleRuns re-arms timers on cold start
- stop() and deleteRun() cancel pending timers
- "autoResuming" event emitted before each attempt

Non-compliant requirements:

- The backoff attempt counter is held only in memory, so the cap resets on process restart and a run can exceed the intended maximum attempts across restarts

Requires further human verification:

- None





**[32](https://github.com/QuintinShaw/pi-dynamic-workflows/issues/32) - Partially compliant**

Compliant requirements:

- src/code-review.ts created with generateCodeReviewWorkflow()
- /code-review registered alongside existing commands
- Input modes for no args, git range, file path, and PR number are implemented
- Model tier routing matches the spec
- 7 parallel finder agents with structured candidate output
- Candidates are deduplicated before verify
- Ranking correctness (A/B/C) before cleanup (D/E/F) before altitude (G) with cap at 10
- Synthesis produces JSON findings and a markdown report

Non-compliant requirements:

- README.md documentation for /code-review is not updated
- Verify pass does not use the verify() stdlib with reviewers: 1, threshold: 0.5; it uses manual agent() calls
- Verifier agents are not given the relevant file(s) context beyond the diff

Requires further human verification:

- Confirm whether the verify() stdlib exists in this codebase; if it does, the manual agent implementation should be replaced



</td></tr>
<tr><td>⏱️&nbsp;<strong>Estimated effort to review</strong>: 3 🔵🔵🔵⚪⚪</td></tr>
<tr><td>🧪&nbsp;<strong>PR contains tests</strong></td></tr>
<tr><td>🔒&nbsp;<strong>No security concerns identified</strong></td></tr>
<tr><td>⚡&nbsp;<strong>Recommended focus areas for review</strong><br><br>

<details><summary><a href='https://github.com/QuintinShaw/pi-dynamic-workflows/pull/46/files#diff-b772f21a1f8e46b7298c49da97c8e42f6038b7a13209db1773713cbb46748f2aR104-R104'><strong>Non-persistent attempt cap</strong></a>

autoResumeAttempts is stored only in a private Map, so the exponential-backoff attempt counter resets to 0 on process restart. A run that has already been auto-resumed several times and re-paused can receive another full set of attempts after a cold start, breaking the MAX_AUTO_RESUME_ATTEMPTS cap and causing repeated resume attempts against an exhausted budget.
</summary>

```typescript
private autoResumeAttempts = new Map<string, number>();
```

</details>

</td></tr>
</table>

gr3enarr0w:
**[Persistent review](https://github.com/QuintinShaw/pi-dynamic-workflows/pull/46#issuecomment-4827470936)** updated to latest commit https://github.com/QuintinShaw/pi-dynamic-workflows/commit/2347f1af4ebdbd025453271833a748ab39ca8d04

QuintinShaw:
This bundles two unrelated features with very different risk profiles, and they're blocking each other: /code-review is close to mergeable, while auto-resume needs another iteration. Please split it — /code-review can land quickly.

**Auto-resume (#27), the things that need rework:**

1. When the timer fires, `resume()` returning `false` (lease held, script missing) is silently swallowed: the attempt is consumed, `autoResuming` was already emitted, no timer is rescheduled — the run stays paused forever. Reschedule on failure or don't consume the attempt.
2. Cold-start re-arm computes the full delay from "now" instead of the remaining time from persisted `updatedAt`. Timers are `unref()`'d, so anyone who restarts pi more often than the hint interval never auto-resumes — #27 explicitly asked for a timer that fires when the reset window elapses.
3. `autoResume: false` isn't persisted: `recoverStaleRuns()` re-arms unconditionally and a re-pause after manual `resume()` gets the default `true`. Same for `stop()` — the persisted `pauseReason` survives, so the next restart re-arms the ghost timer.
4. Please add a floor on the delay (a hint parsing to ~0 can fire before `persistRun` writes the paused state — the callback then sees a non-paused run and drops tracking), and a fake-timer test that actually fires the callback end-to-end; the current backoff test patches `setTimeout` so the fire path is never executed.

**/code-review (#32), minor:** raise `execFileAsync`'s `maxBuffer` (default 1 MB — `gh pr diff` on a big PR throws) and cap/truncate the diff fed into the 7 finder prompts; the issue's file list includes a README update; and either use the existing `verify()` stdlib for the verify phase per the issue spec or note why the hand-rolled verifier is preferable.

QuintinShaw:
Hey @gr3enarr0w — no rush at all, just checking in since it's been a couple weeks. Are you planning to pick up the review notes here? The `/code-review` half is close to mergeable; the auto-resume half is where the rework is (the reschedule/persistence points above). If you're swamped, I'm happy to take it forward from here and keep you credited — just say which you'd prefer.

QuintinShaw:
Heads up — I've split the `/code-review` half of this PR out and landed it separately in #62 (merged), crediting you as co-author. It's the same command with a few maintainer follow-ups on top: diff-capture `maxBuffer` raised to 64MB with a clear over-limit message, a 200k-char truncation guard, and README docs. Closed #32.

This PR stays open for the **auto-resume** half, which still needs the rework from my earlier review (the `resume()`→false swallow, cold-start re-arm using the full delay vs remaining-from-`updatedAt`, the `autoResume:false`+`stop()` ghost re-arm, a delay floor, and the untested timer-fire path). No rush — happy to take that forward too if you'd rather hand it off, otherwise it's all yours. Thanks again for both features.

QuintinShaw:
Hi @gr3enarr0w — an update on this one. The `/code-review` half shipped separately a while back as #62. For the auto-resume half: since it had been open a few weeks and auto-resume-on-usage-limit is something I needed for my own long runs, I implemented it in **#78** (now merged, closing #27).

I want to be straight that your PR directly informed that work — reviewing it is what surfaced the exact edge cases the rewrite had to get right (the swallowed `resume()`→false, cold-start using remaining vs full delay, the un-persisted opt-out, the missing delay floor, the untested timer path). So thank you — it wasn't wasted even though the final code took a different route (a standalone scheduler, to stay decoupled from manager internals).

I think this can be closed now that both halves are covered (#62 + #78), but I'll leave that to you in case you'd rather repurpose it. Really appreciate the contributions — hope to see more.

QuintinShaw:
Closing this out now that both halves have shipped elsewhere:

- /code-review → #62
- auto-resume after usage-limit pause → #78

Your earlier branch still helped a lot — the review notes on it are what forced the edge cases in #78 to get fixed properly. Thanks again for the work here.

PR #49, opened 2026-06-29
fix: share host ModelRegistry with workflow subagents

dizthewize:
**Problem**
Workflow subagents build an isolated `ModelRegistry` from `auth.json` + `models.json`. Providers registered dynamically by extensions (e.g. `pi-ollama-cloud`) are not visible to this isolated registry. As a result, tier/phase routing to `ollama-cloud/*` models silently falls back to the session default (often Anthropic), producing unexpected credit-balance errors and costs.

**Fix**
Thread the host Pi session's `ctx.modelRegistry` through the extension so subagents resolve models against the same registry as the main Pi session:
- `extensions/workflow.ts` passes `ctx.modelRegistry` to `WorkflowManager` via `setModelRegistry()`
- `WorkflowManager` stores it and forwards it into `runWorkflow`
- `runWorkflow` passes it to `WorkflowAgent` via constructor options and per-run `agent()` options
- `WorkflowAgent` reuses the shared registry for both model resolution and the `createAgentSession` call
- Nested `workflow()` calls inherit `modelRegistry` via `...options`
- `listAvailableModelSpecs()` optionally accepts the shared registry so the workflow tool advertises the same models as the host session

The per-run `modelRegistry` option is forwarded explicitly to injected runners primarily for test observability and to keep the option surface consistent with the constructor.

**Verification**
```js
export const meta = { name: 'routing_check', description: 'verify tier routing', phases: [{ title: 'Routing' }] };
const result = await agent('Return JSON {provider, modelId}.', {
  label: 'medium probe',
  tier: 'medium',
  schema: { type: 'object', properties: { provider: { type: 'string' }, modelId: { type: 'string' } }, required: ['provider', 'modelId'] }
});
// result.provider === 'ollama-cloud'
```

**Tests added**
- `WorkflowAgent` reuses an injected `ModelRegistry`
- `runWorkflow` forwards `modelRegistry` to the runner
- `WorkflowManager.setModelRegistry` stores and forwards the registry

**Files changed**
- `src/agent.ts`
- `src/workflow.ts`
- `src/workflow-manager.ts`
- `extensions/workflow.ts`
- `tests/agent.test.ts`
- `tests/workflow-manager.test.ts`

### DIFF
--- a/extensions/workflow.ts
+++ b/extensions/workflow.ts
@@ -54,6 +54,9 @@ export default function extension(pi: ExtensionAPI) {
     // Tell the manager the session's main model so "explore" agents auto-tier
     // down to a lighter same-family sibling (e.g. Claude → Haiku).
     manager.setMainModel(ctx.model ? `${ctx.model.provider}/${ctx.model.id}` : undefined);
+    // Share the host session's model registry so tier/phase routing resolves
+    // extension-registered providers (e.g. ollama-cloud) consistently.
+    manager.setModelRegistry(ctx.modelRegistry);
     // Scope the /workflows history to this session: runs persist on disk across
     // sessions, but the navigator/task panel show only the current session's runs.
     // Switching back to a previous session re-shows that session's runs.
--- a/extensions/workflow.ts
+++ b/extensions/workflow.ts
@@ -47,16 +47,19 @@ export default function extension(pi: ExtensionAPI) {
   let editorInstalled = false;
 
   pi.on("session_start", (_event: unknown, ctx: ExtensionContext) => {
-    const active = pi.getActiveTools();
-    if (!active.includes(workflowTool.name)) {
-      pi.setActiveTools([...active, workflowTool.name]);
-    }
     // Tell the manager the session's main model so "explore" agents auto-tier
     // down to a lighter same-family sibling (e.g. Claude → Haiku).
     manager.setMainModel(ctx.model ? `${ctx.model.provider}/${ctx.model.id}` : undefined);
     // Share the host session's model registry so tier/phase routing resolves
-    // extension-registered providers (e.g. ollama-cloud) consistently.
+    // extension-registered providers (e.g. ollama-cloud) consistently. Set it
+    // before activating the tool: the tool's promptGuidelines read the
+    // manager's registry lazily, so tool-registry refreshes from here on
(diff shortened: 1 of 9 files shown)


QuintinShaw:
Thanks — this is a clean fix and it's wired the right way (constructor path → `WorkflowAgent.getRegistry()` → `createAgentSession`). I verified locally against current main: tsc clean, full suite 750/750 green. Two things before merge:

1. The PR body says the workflow tool now advertises the same models as the host session, but `modelRoutingGuideline()` in `src/workflow-tool.ts` still calls `listAvailableModelSpecs()` with no registry — so the workflow-authoring LLM is still only told about disk-registry models and may never route to `router/...` on its own. Could you thread the registry through there too? Note the registry object is live (router extensions also register on `session_start`), so prefer a lazy lookup over snapshotting at tool-creation time.
2. The per-run `modelRegistry` option passed at `workflow.ts` (~L472) isn't on `AgentRunOptions` and `WorkflowAgent.run()` never reads it — it only survives through the existing `as any` cast. Either add it to the type and honor it, or drop it; as-is it's dead code.

I've approved CI. Once these two land I'll do a verification run with a dynamically-registered provider and merge.

QuintinShaw:
Update: I ran the end-to-end verification early, against the current head. Registered an in-memory provider via `registry.registerProvider(...)` (the same public API pi-model-router uses on `ctx.modelRegistry`), pointed a workflow agent at it — resolves and completes correctly with `setModelRegistry` wired, no fallback warning; the control run without it reproduces the silent session-default fallback from #50. Disk providers unaffected. So runtime-wise this is solid.

Rather than block on the two small items above, I've pushed them to this branch directly (hope you don't mind): the workflow tool's `promptGuidelines` is now a lazy accessor reading `manager.getModelRegistry()` — the SDK re-reads it on every tool-registry refresh, so the advertised model list tracks late-registered providers — and the per-run `modelRegistry` option is now typed on `AgentRunOptions` and honored by `WorkflowAgent.run()` (per-run wins over the constructor's registry), which also let me drop the `as any` cast. Merging once CI is green.

QuintinShaw:
Merged — thanks for tracking this down and for the clean wiring. This closes the main gap for router-style extensions; it'll go out in the next release.

PR #52, opened 2026-07-06
feat: use workflow meta.name as semantic prefix in run IDs

chendpoc:
## What

Workflow run IDs (e.g. `mr90ravb-lnxzrd`) are currently opaque random strings. This PR uses the workflow's `meta.name` (e.g. `profile_sync_upgrade`) as a human-readable prefix, so run IDs become `profile-sync-upgrade-mr90ravb-lnxzrd`.

## How

- Slugify `meta.name` in `startInBackground()` and `createManaged()`
- Prepend slug to the auto-generated `generateRunId()` output
- Fallback to original format when `meta.name` is empty or missing

## What's unchanged

- No API or schema changes — fully backward compatible
- `generateRunId()` signature untouched
- `/workflows status` already shows `workflowName`; this just makes IDs readable too
- Single file: `src/workflow-manager.ts` (+18/-3)

### DIFF
--- a/src/workflow-manager.ts
+++ b/src/workflow-manager.ts
@@ -176,9 +176,16 @@ export class WorkflowManager extends EventEmitter {
     args?: unknown,
     exec: ExecOptions = {},
   ): { runId: string; promise: Promise<WorkflowRunResult> } {
-    const runId = generateRunId();
-    const controller = new AbortController();
     const parsed = parseWorkflowScript(script);
+    const slug = parsed.meta.name
+      ? parsed.meta.name
+          .toLowerCase()
+          .replace(/[^a-z0-9]+/g, "-")
+          .replace(/^-+|-+$/g, "")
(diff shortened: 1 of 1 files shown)


QuintinShaw:
Thanks for this — clean, well-scoped, and it holds up under adversarial input: I tested the sanitization against deliberately hostile names (`../weird name: profile/sync!!`) in both code review and a live run, and it collapses to a safe slug with uniqueness, resume, and old-id compatibility all intact. A few PRs landed since you opened this, so I rebased it onto latest `main` and I'm merging now. Two entirely optional bits of future polish (nothing you need to do): the slugify block is duplicated between `startInBackground()` and `createManaged()` (could fold into a small helper), and the description's "fallback when `meta.name` is empty" path is effectively dead since the parser already guarantees a non-empty name. Really nice first contribution.

PR #53, opened 2026-07-07
fix(agent-registry): align user-level agent dir and tools parsing with pi-coding-agent conventions

happyTonakai:
## Summary

Aligns `loadAgentRegistry` and the `tools` / `disallowedTools` parser with the conventions actually used by pi-coding-agent. Today, an `agentType` reference silently falls back to "unknown agentType" for any user whose agent markdowns live at the standard pi-coding-agent location (`~/.pi/agent/agents/*.md`), because two related bugs in `agent-registry.ts` prevent the files from being bound correctly.

## Problem

**1. Wrong user-level scan path**

The user-level directory was scanned at `~/.pi/agents/`, but pi-coding-agent's built-in agent discovery and the official subagent extension example both scan `~/.pi/agent/agents/` (i.e. `getAgentDir() + "agents"`).

For users with markdowns at the standard location, the registry came back empty and every `agentType` reference logged the fallback warning.

**2. `tools` / `disallowedTools` parsing only accepts YAML list form**

```ts
function toStringArray(value: unknown): string[] | undefined {
  if (!Array.isArray(value)) return undefined;  // strings always rejected
  ...
}
```

But `parseFrontmatter` (the standard frontmatter parser from pi-coding-agent) returns `tools` as a comma-separated **string** for `tools: read, grep, find, ls`, not an array. The official subagent example just calls `.split(",")` on it. QuintinShaw's own tests cover the YAML list form (`tools: [read, grep]`) but never the string form, so the bug slipped through.

## Fix

**1. Use `getAgentDir()`** for the user-level directory:

```diff
- const userDir = opts?.userDir ?? join(homedir(), AGENTS_DIR);
+ const userDir = opts?.userDir ?? join(getAgentDir(), "agents");
```

This matches pi-coding-agent's standard path and also honors the `PI_CODING_AGENT_DIR` env override. Project-level resolution (`cwd + ".pi/agents/"`) is unchanged.

**2. Extend `toStringArray()`** to also split a comma-separated string, matching the format used by the official subagent example:

```diff
 function toStringArray(value: unknown): string[] | undefined {
+  if (value == null) return undefined;
+  if (Array.isArray(value)) {
+    const arr = value.filter(...).map(...);
+    return arr.length ? arr : undefined;
+  }
+  if (typeof value === "string" && value.trim().length > 0) {
+    return value.split(",").map((s) => s.trim()).filter(Boolean);
+  }
   return undefined;
 }
```

The existing array-form path is preserved so the project's own tests (still passing) keep working.

## Verification

- `tsc --noEmit`: clean
- `npx tsx --test tests/*.test.ts`: **787 / 787 passing**
- `tests/agent-registry.test.ts`: 25 / 25 passing (including the existing array-form cases)
- End-to-end: with markdowns at the standard `~/.pi/agent/agents/*.md` location, both `model` and `tools` now resolve correctly and `source` is reported as `"user"`.

### End-to-end output

```
=== Result: found 2 agentTypes ===
  - code-reviewer
    source: user
    model:  <provider>/<model-id>     # from the markdown frontmatter
    tools:  ["read","grep","find","ls","bash","codegraph_affected"]
  - scout
    source: user
    model:  <provider>/<model-id>     # from the markdown frontmatter
    tools:  ["read","grep","find","ls"]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ The full guide — every global, agent option, `agentType` definitions, structur
 | --- | --- |
 | `tier` | `"small"` \| `"medium"` \| `"big"` — coarse model routing (configure via `/workflows-models`; tiers may store `provider/modelId:thinking`). |
 | `model` | Exact `provider/modelId` or `provider/modelId:thinking` (always wins over `tier`). |
-| `agentType` | A named definition (`.pi/agents/<name>.md`) binding tools + model + role prompt. |
+| `agentType` | A named definition (`.pi/agents/<name>.md` project-level, or `~/.pi/agent/agents/<name>.md` user-level — `~/.pi/agents/<name>.md` still works as a deprecated fallback) binding tools + model + role prompt. |
 | `isolation: "worktree"` | Run in a throwaway git worktree for conflict-free parallel edits. |
 | `schema` | JSON Schema → the subagent returns a validated object. |
 | `label` / `phase` / `timeoutMs` | Display label / phase override / optional per-agent hard timeout. Omit `timeoutMs` for no hard timeout. |
--- a/src/agent-registry.ts
+++ b/src/agent-registry.ts
@@ -6,8 +6,13 @@
  *   agent("audit this dir", { agentType: "security-auditor" })
  *
  * Definitions live as Markdown files under `.pi/agents/*.md` (project, cwd-relative)
- * and `~/.pi/agents/*.md` (user). Frontmatter binds the subagent's tools, model,
- * and a body prompt; project definitions win on a name collision. This mirrors
+ * and `~/.pi/agent/agents/*.md` (user — `getAgentDir() + "agents"`, honoring the
+ * `PI_CODING_AGENT_DIR` override), matching pi-coding-agent's own built-in agent
+ * discovery convention. The legacy `~/.pi/agents/*.md` location is still scanned as
+ * a deprecated fallback (with a one-time warning) so users who followed this repo's
+ * earlier docs are not silently broken; the new location wins on a name collision.
+ * Frontmatter binds the subagent's tools, model, and a body prompt; project
+ * definitions win over both user-level locations on a name collision. This mirrors
  * Claude Code's `.claude/agents` registry: agentType is a real binding of
  * tools+model+system-prompt, not a prose hint.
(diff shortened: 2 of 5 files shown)


QuintinShaw:
Thanks — this is a well-researched fix, and the core diagnosis is correct. I verified it against the actual `@earendil-works/pi-coding-agent` SDK (including the official subagent example): the user-level dir really is `~/.pi/agent/agents/` and `tools:` really is comma-separated, and I confirmed the pre-fix code returns an empty registry on that fixture — so the bug is real, not cosmetic.

Three things before merge:
1. A few doc comments still point at the old path, including in the file you touched — `src/agent-registry.ts:8-9`, `src/config.ts:48`, `src/workflow.ts:72`. I'm happy to push those doc updates to your branch myself if you'd prefer.
2. Could you add a unit test for the comma-separated `tools`/`disallowedTools` string form (and ideally one for default `userDir` resolution via `getAgentDir()`)? That's the part I'd really like from you, since it locks in the behavior.
3. One decision for us: users who followed this repo's earlier docs and put agent files at `~/.pi/agents/` will have them silently stop resolving after this merges. How would you feel about also scanning the old location as a deprecated fallback (with a one-time warning), versus a clean cutover plus a release note? Genuinely your call — I'm fine either way, just want it to be deliberate rather than silent.

Really solid first contribution.

QuintinShaw:
Landed this — thanks, your diagnosis and fix were spot on. I verified against the real `@earendil-works/pi-coding-agent` (the user dir really is `~/.pi/agent/agents/` and `tools:` really is comma-separated), and pushed a few follow-ups to your branch to round it out: (1) the stale doc comments (`agent-registry.ts`, `config.ts`, `workflow.ts`) now describe the new path; (2) added tests for the comma-string form and default `getAgentDir()` resolution; and (3) for the migration concern I raised — rather than a hard cutover that would silently break anyone who followed the old docs, it now **dual-scans**: new `~/.pi/agent/agents/` takes precedence, but the old `~/.pi/agents/` is still read as a deprecated fallback with a one-time warning pointing people to the new location. 815/815 green, real-pi verified (old→resolves+warns, new→shadows). Really appreciate the fix.

PR #51, opened 2026-07-02
feat: support Pi CLI-style thinking suffixes in workflow model specs

Whamp:
## What

Workflow model specs and tiers now accept an optional Pi CLI-style thinking suffix, e.g. `openai-codex/gpt-5.5:xhigh`, parsed with the same grammar as `pi --model`. When an agent resolves such a spec, the matched model is used and the thinking level is forwarded to the created subagent session.

```json
{
  "tiers": {
    "small": "openai-codex/gpt-5.4-mini:low",
    "medium": "openai-codex/gpt-5.4:medium",
    "big": "openai-codex/gpt-5.5:xhigh"
  }
}
```

`/workflows-models` now offers a thinking-level picker (`off`/`minimal`/`low`/`medium`/`high`/`xhigh`, or session default) after the model is chosen, and preselects the base model when the stored tier already carries a suffix.

## Why

The CLI already lets users dial thinking per model. Workflows routed by `opts.tier` / `opts.model` could not, so tier-bound subagents ran at the session default thinking level regardless of intent. This makes the workflow grammar match the rest of Pi.

## Changes

- **`src/model-spec.ts`** (new): `canonicalModelSpec`, `splitModelSpecThinking`, `formatModelSpecWithThinking`, `isThinkingLevel`, `THINKING_LEVELS`, and registry-aware `resolveModelSpecWithThinking`. Includes `fast-check` property tests for the round-trip.
- **`src/agent.ts`**: `run()` resolves the spec via `resolveModelSpecWithThinking` and forwards the parsed `thinkingLevel` to `createAgentSession`. `listAvailableModelSpecs` canonicalizes its output so a tier with a suffix round-trips through the picker.
- **`src/workflow-manager.ts` / `extensions/workflow.ts` / `src/workflow.ts`**: thread the host session's `ModelRegistry` into each `agent()` run so dynamic/authed providers resolve like the parent session (composes with #49).
- **`src/workflows-models-command.ts`**: second-stage thinking picker in the tier editor.
- **`src/index.ts`**: export the new model-spec API.
- **`README.md`**: document the `provider/modelId:thinking` format in the agent-options table and the `model-tiers.json` paragraph.

## Backward compatibility

All new options are additive. A spec without a suffix resolves identically to before. No settings migration is required; existing `model-tiers.json` values keep working.

## Verification

- `npm test` green: Biome check + `tsc` build + **767** unit tests.
[...]

PR #54, opened 2026-07-08
feat: add persistAgentSessions setting for durable subagent transcripts

codesoda:
## Problem

Workflow subagents run with a hardcoded `SessionManager.inMemory()` (`src/agent.ts`), so no `.jsonl` session files are ever written. The only surviving record is the compacted history embedded in the run JSON (`~/.pi/workflows/projects/<key>/runs/<runId>.json`), which `src/agent-history.ts` truncates hard (max 40 entries, 2,000 chars/entry, 20,000 chars total per agent). Full transcripts are unrecoverable — e.g. a 35-agent run that opened 8 PRs leaves no way to audit what any agent actually did.

## What this adds

A new opt-in setting `persistAgentSessions` (boolean, **default `false`** — current behavior is preserved exactly):

- When `true`, each `agent()` call creates a file-backed `SessionManager` so the full transcript lands in the standard sessions dir (`~/.pi/agent/sessions/<encoded-cwd>/`), inspectable with existing session tooling (`/resume` etc.).
- Same precedence as other settings: project-level settings override global (`loadWorkflowSettings({ cwd })` already merges that way).
- Sessions are keyed by the runner's **project cwd** (`this.cwd`), not the per-call `runCwd`, so agents working in temporary git worktrees still group under the project's session dir instead of scattering across short-lived worktree paths.
- Persisted sessions are named `workflow:<runId> <agent label>` via `appendSessionInfo`, so they're identifiable in session pickers.
- Injected `session.sessionManager` overrides (tests/embedders) still win — the `...sessionOptions` spread ordering is preserved.

## Changes

- `src/workflow-settings.ts` — `persistAgentSessions?: boolean` on `WorkflowSettings` + boolean-only normalization
- `src/agent.ts` — `WorkflowAgentOptions.persistAgentSessions`, `createSessionManager()` seam, `AgentRunOptions.sessionName`, best-effort session naming
- `src/workflow.ts` — passes `sessionName: workflow:<runId> <label>` per agent call (the flag itself rides along since `WorkflowRunOptions extends WorkflowAgentOptions`; nested runs inherit via `...options`)
- `src/workflow-manager.ts` — `WorkflowManagerOptions.persistAgentSessions`, forwarded to `runWorkflow`
- `extensions/workflow.ts` — wires `settings.persistAgentSessions` into the manager
- `README.md` — settings docs incl. default, where files land, and the picker-clutter caveat for large fan-outs
- Tests: settings parsing (booleans only, absent default, project-over-global precedence), agent session-manager selection (in-memory default, file-backed keyed by project cwd), manager plumb-through + sessionName format

`npm test` (check + build + 795 unit tests) passes.

## Out of scope / possible follow-ups

- Per-run override on the workflow tool schema (e.g. a `persistAgentSessions` tool-call param) — settings-level only in v1
- Raising `agent-history.ts` JSON retention limits / cleanup for large fan-outs


QuintinShaw:
Really clean, well-scoped feature — and nice that it reuses pi's existing `SessionManager` rather than inventing a writer. I verified it end to end: it's off by default (unset behavior is byte-identical to before), filenames are per-call uuidv7 so parallel agents can't collide, and with a real subagent run, enabling it writes a valid `.jsonl` transcript named `workflow:<runId> <label>` while disabled writes nothing.

Two asks, plus a rebase I'll handle:
1. Failure tolerance: you already wrap the session-naming step (`appendSessionInfo`) in try/catch — could you extend the same care to the session write itself? Right now a mid-run write failure (disk full / permissions) would throw and abort that subagent's run, whereas degrading to an in-memory session would be safer. Workflow subagents never touched disk before this flag, so it's a genuinely new failure surface worth guarding.
2. Could you add one sentence to the README caveat noting the persisted transcript is full/untruncated (unlike the compacted run JSON), so anything a subagent reads into context — including secrets — lands on disk when this is enabled?
3. No action needed here — there's a trivial one-method conflict with `main` (a `resolveModel()` leftover that #51 removed); I'll take care of the rebase so you can focus on the two items above.

Strong contribution — thanks for it.

QuintinShaw:
Landed this — clean, well-scoped, and nice reuse of pi's existing `SessionManager`. I verified it end to end (off by default, byte-identical when unset, per-call uuidv7 filenames, real `.jsonl` written when enabled) and pushed the two follow-ups to your branch: (1) **failure tolerance** — since the SDK's actual write happens lazily inside `session.prompt()` and can't be caught from our side, `createSessionManager()` now write-probes the session dir right after creation and, on any failure (permissions/disk-full), logs a warning and degrades that agent to an in-memory session instead of aborting the run (per-agent, doesn't affect siblings); (2) the README now notes the transcript is full/untruncated so secrets a subagent reads land on disk when enabled. Added a test that forces the failure path. 819/819 green, real-pi verified (enabled writes / unset none / failure degrades). Thanks for the feature.

PR #56, opened 2026-07-09
feat(workflows): add thinking levels and live token progress

kevinpita:
Adds workflow thinking-level support and richer live progress reporting.

 - Extends workflow model tiers to store both model and optional thinkingLevel, while preserving backward compatibility with legacy string-only configs.

 - Adds opts.thinkingLevel to agent() and applies tier/default thinking levels to subagent sessions.
 - Updates /workflows-models so each tier can edit model + thinking level (inherit, off, minimal, low, medium, high, xhigh).

 - Adds live token progress for running agents, including estimated tokens while streaming/thinking and finalized provider usage when available.
 - Enhances the workflow panel/navigator to show per-agent tokens, model + thinking level, aggregate cost, and live tok/s in detailed mode.

 - Updates README and expands unit coverage for tier normalization, thinking-level routing, token progress, persistence, and UI rendering.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ The same model — on Pi, plus the production pieces a real run needs:
                             switch the live panel between the compact one-liner and the detailed
                             per-phase/per-agent view (with tokens, cost, and a live tok/s rate)
 /workflows-progress-max <N> cap agents shown per phase in detailed mode (1-1000, default 8)
-/workflows-models           map the small / medium / big tiers to real models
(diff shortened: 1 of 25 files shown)


Whamp:
how is this different from #51 ?

kevinpita:
> how is this different from #51 ?

Actually, it isn’t. Sorry for the duplicate, and nice work on your PR!

PR #57, opened 2026-07-09
feat: show a fresh vs cached token split in the workflow displays

vssnake:
## What

The workflow token displays show a single scalar per agent — `input + output + cacheRead + cacheWrite` — which mixes real spend with cheap cache reads. For a ~95%-cached Opus agent that reads as huge (e.g. "3.0M tok" when only ~89k was actually fresh); for a non-caching provider it's 100% fresh. Same unit, opposite meaning — it's easy to misread which agents actually cost you.

This surfaces the split the SDK already computes (`AgentUsage`), which the workflow layer was collapsing to a scalar.

**Before**
```
[1] ✓ paper-generation   3.0M tok    · claude-opus-4.8
[2] ✓ numeric-auditor    6.3M tok    · glm-5-2
```

**After**
```
[1] ✓ paper-generation   89K fresh · 3.0M cache   · claude-opus-4.8
[2] ✓ numeric-auditor    6.3M fresh               · glm-5-2
```
Delivery line: `✓ … finished (11 agents · 18.0M fresh · 6.0M cache · $6.70 · 51.9s)`.

`fresh = input + output` (real spend), `cache = cacheRead` (cheap reuse). The cache segment is omitted when there were no cache reads, so a non-caching run reads cleanly as `6.3M fresh`. `cacheWrite` is excluded from both counts (its cost is reflected in `$cost` on the delivery line).

## How

- Carry the per-agent `AgentUsage` (already captured via `onUsage`) through the existing `onAgentEnd` event into `WorkflowAgentSnapshot.tokenUsage` — it was being dropped.
- Render the split across **all** per-agent token surfaces for consistency: the live detailed panel (`task-panel.ts`), the `/workflows` navigator (`workflow-ui.ts`), the status/print command (`display.ts`), and the background-run delivery line.
- Backward compatible: agents without a `tokenUsage` breakdown fall back to the existing `N tok` scalar.

`feat:` because it adds public surface (`WorkflowAgentSnapshot.tokenUsage` and the `onAgentEnd` event's `tokenUsage`). Display-only — nothing changes about what's sent to models or how much is spent.

## Verification

`npm test` green (biome check + tsc build + 790 unit tests). Added coverage in `task-panel.test.ts` (per-agent split, delivery-line cache+cost, cache-omitted-when-`cacheRead=0`) and `workflow-display.test.ts` (status-command split).

**End-to-end against a real Pi subagent session** (per CONTRIBUTING's "token accounting" rule): verified on a live install carrying the same change. Real runs on `forge/glm-5-2` (parallel fan-out) + `github-copilot/claude-opus-4.8` confirm the per-agent `tokenUsage` populates from the real `createAgentSession` — the run-level aggregate was `input+output ≈ 18.0M`, `cacheRead ≈ 6.03M`, `cacheWrite ≈ 234k`, and the `cacheRead` is entirely attributable to the copilot/Opus agents (the forge/GLM provider reports no cache reads). So the panel renders the Opus agents with a large `cache` segment and the GLM agents as `fresh`-only, exactly as intended. Happy to share the raw run stats if useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ return await agent('Synthesize and double-check these findings:\n' + findings.jo
 - **Journaled resume** — an interrupted run replays finished agents from a journal (no re-run, no tokens) and runs only what's left or what you changed.
 - **Git worktree isolation** — `isolation: "worktree"` gives an agent its own branch, so parallel agents can edit the same files without clobbering each other.
 - **Real token & cost accounting** — read from each subagent's session, not estimated. Runs have no default token cap; `tokenBudget`, phase budgets, and `budget` let you add explicit gates when you want them.
-- **Background by default** — the turn ends right away, a live "Workflows running" panel tracks runs, and each result is delivered back so the conversation auto-continues when it finishes. The panel is compact by default; `/workflows-progress detailed` expands it inline to per-phase/per-agent rows with tokens, cost, and a live tok/s rate (so a stalled agent shows as 0 tok/s) — no need to open `/workflows`.
+- **Background by default** — the turn ends right away, a live "Workflows running" panel tracks runs, and each result is delivered back so the conversation auto-continues when it finishes. The panel is compact by default; `/workflows-progress detailed` expands it inline to per-phase/per-agent rows with a fresh/cache token split, cost, and a live tok/s rate (so a stalled agent shows as 0 tok/s) — no need to open `/workflows`.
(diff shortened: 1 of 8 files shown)


vssnake:
Closing this for now. In wider local testing the fresh/cache split only renders for providers whose session surfaces per-agent usage via getSessionStats(); for agents running through a provider that doesn't report token usage (e.g. a vLLM gateway), onUsage never fires and the panel falls back to the scalar `N tok`. That makes the feature inconsistent across setups. I'd rather revisit with a more robust usage source before merging. Thanks for the thorough review!

PR #58, opened 2026-07-09
feat: point delivered results at the full persisted file with configurable truncation

### LINKED ISSUE #55: Delivered workflow result is silently truncated at 400 chars with no pointer to the full result
When a background run finishes, the result delivered to the conversation
(`summarizeResult` in `deliverText`) is cut to 400 chars whenever the return
value is an object without a `verdict`/`report`/`summary` string field —
silently, with no hint of how much was dropped or where the full result lives:

https://github.com/QuintinShaw/pi-dynamic-workflows/blob/e2ea9633a88ffc6f144e0ebd3c83a2ecd5545efc/src/task-panel.ts#L51-L62

The full result is persisted at `~/.pi/workflows/projects/<proj>/runs/<id>.json`,
but the delivered message never points there, so the tail just vanishes from the
assistant's view. Easy to hit: returning `{ ok, changed, applied: { note } }`
with a normal-length `note` already blows past 400 chars and gets cut mid-field.

**Suggestion (backward-compatible):** always append the persisted result path in
`deliverText` — even when nothing is truncated or a `summary` field exists — and
note the dropped size on truncation, e.g.
`…(truncated 3.2KB — full result: <path>)`. Optionally expose the 400 threshold
as a `~/.pi/workflows/settings.json` knob.

Happy to send a PR if the direction sounds good.

Qusic:
## What & why

Fixes #55. When a background workflow run finishes, `summarizeResult` / `deliverText`
(`src/task-panel.ts`) truncated the delivered result to 400 chars in the JSON-dump
branch — silently, with no size and no pointer to where the full result is persisted.

This PR:

- **Always appends a pointer** to the persisted result
  (`↳ Full result: ~/.pi/workflows/projects/<project>/runs/<id>.json`), even when
  nothing was truncated or a `verdict` / `report` / `summary` field is used.
- **Reports the dropped size** on truncation, e.g. `…(truncated 3.2 KB)`.
- **Makes the threshold configurable** via a new optional `deliveredResultMaxChars`
  setting in `~/.pi/workflows/settings.json` (default `400`, so existing behavior is
  unchanged). Only the JSON-dump fallback honors it; string / verdict / report /
  summary results are never truncated apart from the appended pointer.

## Notes

- Documented in `README.md` (settings paragraph, next to the existing knobs).
- `npm test` is green (biome + tsc + unit). Added unit tests for the pointer, the
  dropped-size marker, the configurable threshold, and settings normalization.
- This only affects the post-completion **delivery text**, not how agents actually
  run (retries / timeouts / routing / tokens / concurrency / resume), so a
  real-provider E2E run is not required — fake-agent unit tests cover it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ For larger or flakier fan-outs, the `workflow` tool also accepts `concurrency` (
 
 The live "Workflows running" panel is configured in the same `~/.pi/workflows/settings.json`: `"progressPanelMode"` is `"compact"` (default, one line per run) or `"detailed"` (per-phase/per-agent rows with tokens, cost, and a live tok/s rate), and `"progressPanelMaxAgents"` (default `8`, range `1`–`1000`) caps how many agents each phase shows in detailed mode before a `… N earlier agents` line. Toggle them live with `/workflows-progress compact|detailed` and `/workflows-progress-max <N>` — changes take effect on the next render without a restart.
 
+When a background run finishes, its result is delivered back into the conversation with a `↳ Full result: <path>` pointer to the persisted `~/.pi/workflows/projects/<project>/runs/<id>.json`, so nothing is lost even when the summary is shortened. Only the JSON-dump fallback (a result object without a `verdict`/`report`/`summary` string field) is truncated — at `"deliveredResultMaxChars"` characters (default `400`) in the same `~/.pi/workflows/settings.json` — and the dropped size is shown inline, e.g. `…(truncated 3.2 KB)`.
+
 Workflows run in a Node `vm` sandbox; `Date.now()`, `Math.random()`, `new Date()`, and `require`/`import`/`fs`/network are unavailable, so runs stay reproducible — which is what makes resume reliable.
 
 ## Default tier assignment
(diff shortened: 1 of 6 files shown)


QuintinShaw:
Thanks — thorough, well-tested first PR, and it matches the direction from #55 exactly. I verified the pointer resolves to the real `<runsDir>/<runId>.json` (confirmed it exists on disk across large/small/string results), the dropped-size reporting is accurate (byte-length, UTF-8 safe), the full-passthrough branches are untouched, and it degrades gracefully when persistence isn't available. The `deliveredResultMaxChars` knob is validated and defaults to 400, so existing behavior is preserved. 808/808 green, no conflicts — I rebased onto current `main` and I'm merging now. Two tiny optional notes for later (nothing to do): `emit("complete")` fires just before the final `persistRun()` write, so the pointer is computed a hair before it lands — harmless, but worth a comment if that ordering ever changes; and `persistedResultPath()` assumes persistence is always on (no `existsSync` guard), which holds today. Really nice work — thanks for both the issue and the fix.

PR #59, opened 2026-07-10
feat: add 'max' thinking level support

tcf909:
Updates `THINKING_LEVELS` and documentation to support the `max` thinking level available in pi-coding-agent. This prevents the parser from rejecting `:max` suffixes in workflow tier configurations.

### DIFF
--- a/src/model-spec.ts
+++ b/src/model-spec.ts
@@ -1,7 +1,7 @@
 import type { Api, Model } from "@earendil-works/pi-ai";
 import type { ModelRegistry } from "@earendil-works/pi-coding-agent";
 
-export const THINKING_LEVELS = ["off", "minimal", "low", "medium", "high", "xhigh"] as const;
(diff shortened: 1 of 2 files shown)


QuintinShaw:
Thanks for this, and for catching the missing `max` level so quickly — your parser fix was correct as far as it goes. Whamp's #61 (filed alongside yours) converged on the same root cause but also traced it through the SDK: on the currently-pinned pi-coding-agent, a parsed `:max` gets silently clamped back to `off` at session creation, so `max` only truly takes effect after bumping to pi-ai/pi-coding-agent ≥0.80.6 (which #61 does, with tests and the picker). Since #61 covers the full chain, I'm going with that one — and I carried your `workflow-tool.ts` guideline line over into it (credited), so your contribution isn't lost. Really appreciate the fix, and hope to see more PRs from you.

QuintinShaw:
Closing in favor of #61, which lands the same `max` fix plus the required SDK bump — details above. Thanks again, @tcf909.

PR #61, opened 2026-07-11
feat: support Pi's max workflow thinking level

### LINKED ISSUE #60: Support Pi's max thinking level in workflow model specs
## Summary

Pi core now supports the `max` thinking level, but pi-dynamic-workflows 2.12.1 defines:

```ts
export const THINKING_LEVELS = ["off", "minimal", "low", "medium", "high", "xhigh"] as const;
```

As a result, a tier configured as `openai-codex/gpt-5.6-sol:max` is not parsed as model + thinking level. The resolver treats `:max` as part of a synthetic model ID. Workflow subagents can then return empty/null output instead of using the intended model.

## Minimal reproduction

Using `resolveModelSpecWithThinking` with a registry containing `openai-codex/gpt-5.6-sol`:

```text
max   -> modelId=gpt-5.6-sol:max, thinkingLevel=undefined
xhigh -> modelId=gpt-5.6-sol,     thinkingLevel=xhigh
```

The `max` result also warns that `gpt-5.6-sol:max` was not found and that a custom model ID is being used.

## Expected behavior

- Add `max` to supported workflow thinking levels so the resolved model remains `gpt-5.6-sol` with `thinkingLevel=max`.
- Prefer failing closed for an unknown suffix rather than constructing a provider model ID containing that suffix.
- Add resolver and `/workflows-models` picker coverage for `max`.

## Environment

- `@quintinshaw/pi-dynamic-workflows`: 2.12.1 (current npm latest)
- Pi core model docs list `off`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`.

Whamp:
## Summary

- add `max` to workflow thinking levels
- expose `max` in `/workflows-models`
- cover `max` in the resolver and picker tests
- require Pi coding-agent and TUI 0.80.6 or newer

## Pi minimum version

The first test run used the lockfile's Pi 0.78.0 packages. That release does not include `max` in `ThinkingLevel`, so TypeScript rejected the resolver output. Pi 0.80.6 includes the type, but updating only `pi-ai` and `pi-coding-agent` produced incompatible TUI types. Updating `pi-tui` to the same release fixed the dependency graph.

This PR records `>=0.80.6` for the coding-agent and TUI peer dependencies. It also moves the faux-provider integration helper to `@earendil-works/pi-ai/compat`, where Pi 0.80.6 exposes that test API.

## Tests

- `npm test` — 810 passed
- resolver regression for `openai-codex/gpt-5.6-sol:max`
- `/workflows-models` picker regression for `max`

Closes #60.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Workflow state is stored under `~/.pi/workflows` so projects do not accumulate e
 }
 ```
 
-Use `/workflows-models` to edit these in the TUI: choose the base model first, then choose `off`, `minimal`, `low`, `medium`, `high`, `xhigh`, or the session default.
+Use `/workflows-models` to edit these in the TUI: choose the base model first, then choose `off`, `minimal`, `low`, `medium`, `high`, `xhigh`, `max`, or the session default.
 
 To avoid accidental keyword triggers, configure a custom trigger word in `~/.pi/workflows/settings.json`:
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,8 +22,8 @@
         "typescript": "latest"
       },
       "peerDependencies": {
-        "@earendil-works/pi-coding-agent": "*",
-        "@earendil-works/pi-tui": "*",
+        "@earendil-works/pi-coding-agent": ">=0.80.6",
+        "@earendil-works/pi-tui": ">=0.80.6",
         "typebox": "*"
       }
     },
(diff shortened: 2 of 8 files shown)


QuintinShaw:
Verified end to end. The key thing your PR gets right that a parser-only change misses: on the current SDK (pi-ai 0.78.0), `clampThinkingLevel(model, "max")` silently returns `off`, so `:max` would parse but run with thinking disabled. Your bump to 0.80.6 is what makes `max` actually take effect (deepseek's top tier moved from `xhigh` to `max`), and existing `:xhigh` callers still get the best tier (it now clamps up). Clean bump: 810/810 tests including the #26 usage-limit guards, tsc clean, npm audit actually improved, and a real `deepseek/…:max` run completed. The `/compat` test-import change is the right adaptation (registerFauxProvider moved there in 0.80.6). One thing — your branch didn't update the `:max` example in `src/workflow-tool.ts`'s `modelRoutingGuideline` (tcf909's #59 had that line), so I folded it in so nothing's lost. Merging now — thanks, and thanks for filing #60 too.

PR #64, opened 2026-07-12
fix(editor): support OMP editor constructor

edde746:
## Summary

Fix the launch crash when this extension is loaded by OMP 16.4.6.

Legacy Pi's `CustomEditor` expects `(tui, theme, keybindings)`, while OMP's `CustomEditor` inherits the newer `Editor(theme)` constructor. Passing the legacy argument list under OMP stored the TUI instance as the editor theme, so the initial render failed while reading `theme.symbols.boxRound`.

The editor now selects the constructor argument layout from the base editor's required parameter count. It also retains the host TUI explicitly for render requests because OMP's newer editor does not expose the legacy `tui` field.

## Compatibility

- Preserves the legacy Pi constructor path.
- Supports OMP's theme-only editor base.
- Does not change the extension's public API or workflow behavior.

## Verification

- `npm run build`
- `npx tsx --test tests/workflow-editor.test.ts`
- `npx biome check src/workflow-editor.ts`
- `omp --no-session -e extensions/workflow.ts`

All automated checks passed. Before the change, the OMP smoke test crashed during initial rendering with `TypeError: undefined is not an object (evaluating 'this.#j.symbols.boxRound')`. After the change, OMP rendered the interactive editor and remained running until the deliberate smoke-test timeout.

### DIFF
--- a/src/workflow-editor.ts
+++ b/src/workflow-editor.ts
@@ -170,6 +170,26 @@ function isBackspace(data: string): boolean {
 }
 
 /**
+ * Pi's CustomEditor still forwards `(tui, theme, keybindings)` to the legacy
+ * pi-tui Editor constructor. OMP's CustomEditor has no constructor and extends
+ * the newer `Editor(theme)`. Select the argument layout from the base Editor's
+ * required parameter count so one extension can run in either host.
+ */
+function customEditorConstructorArgs(
+  tui: TUI,
+  theme: EditorTheme,
+  keybindings: ConstructorParameters<typeof CustomEditor>[2],
+): ConstructorParameters<typeof CustomEditor> {
+  const baseEditor = Object.getPrototypeOf(CustomEditor) as { readonly length: number };
+  if (baseEditor.length === 1) {
+    // The installed Pi types describe the legacy signature; OMP provides the
+    // runtime-only `(theme, keybindings)` signature.
+    return [theme, keybindings] as unknown as ConstructorParameters<typeof CustomEditor>;
+  }
+  return [tui, theme, keybindings];
+}
+
+/**
  * Editor that paints the trigger words and owns the on/off toggle. Reads/writes
  * `state.active` so the extension's `input` handler can decide whether to force a
  * workflow at submit time.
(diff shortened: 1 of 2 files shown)


QuintinShaw:
This looks right, and importantly the legacy Pi path is verifiably unregressed — I ran the editor construct/render tests against the installed legacy base and they pass. Nice touch that reading the base `Editor`'s param count (`.length`) is actually more robust than an identity check here: `pi-coding-agent` bundles its own nested `pi-tui`, so `=== Editor` would silently miss. Two small things + one confirmation before merge:

1. The `length === 1 ? OMP : legacy` branch treats "anything but 1" as legacy implicitly — could you add a brief comment (or a dev assertion/warning) marking that as a deliberate default, so an unexpected future arity can't silently reintroduce the crash?
2. Could you add a small unit test for the arg-layout helper using a fake 1-arg base class? OMP isn't installed on our side, so that closes the only coverage gap without needing OMP.
3. Since we can't run OMP here: could you paste what `Object.getPrototypeOf(CustomEditor).length` returns on your OMP 16.4.6 install? That confirms the `=== 1` detection matches reality.

Thanks — genuinely solid first contribution.

QuintinShaw:
Thanks @edde746 — a well-reasoned fix, and reading the base Editor's arity at runtime is exactly right here: pi-coding-agent bundles its own nested pi-tui, so an identity check (`=== Editor`) would silently miss.

I verified against the published pi-tui@0.80.6 that the base Editor takes 2 args, so this is a clean no-op for vanilla Pi. To keep it moving I addressed my earlier review notes myself: marked the non-1 arity as intentionally legacy with a one-time warn for any unknown arity, exported the helper, and added unit tests for both arg layouts. I also layered a defensive try/catch around `super.render()` (commit 6097922) for issue #72, where OMP's bundled build crashes on a private field our published pi-tui doesn't even have — so even if the constructor-arg theory isn't the whole story there, the app degrades to a plain editor instead of hard-crashing at launch.

Merging now; ships in 2.13.2. Thanks for the sharp diagnosis.

Issue #65, opened 2026-07-14
Reduce always-on workflow prompt size (model catalogue adds ~5k tokens)


Whamp:
## Problem

The extension adds a large amount of always-on context, even when `workflow` is not used. In my setup, Pi's context report attributes about **6,500 tokens** to `pi-dynamic-workflows`; the full model catalogue accounts for about **5,000 tokens** by itself.

I also measured the rendered payload with 363 available models and six `agentType` definitions:

- workflow extension system-prompt contribution: **20,485 UTF-8 bytes**
- dynamic model + `agentType` catalogues: **14,861 bytes**
- provider-visible workflow tool definition: **2,204 bytes**

Every request pays this context cost. Catalogue changes can also disturb prompt-prefix caching.

## Cause

[`modelRoutingGuideline()` enumerates every available model](https://github.com/QuintinShaw/pi-dynamic-workflows/blob/98cbe34a3adc7803217b13adae3e896a5ab01c70/src/workflow-tool.ts#L32-L47), then [`promptGuidelines` adds the generated string to the system prompt](https://github.com/QuintinShaw/pi-dynamic-workflows/blob/98cbe34a3adc7803217b13adae3e896a5ab01c70/src/workflow-tool.ts#L171-L197).

The remaining tool description, parameter descriptions, and 24 prompt guidelines also repeat schema facts or document advanced features that are not needed on every turn.

## Proposed PR

Would you be open to a PR that:

1. removes the full model catalogue from `promptGuidelines`;
2. routes normal work through `small` / `medium` / `big`, using an exact model only when the user supplies one;
3. optionally removes the always-on `agentType` catalogue as well;
4. reduces the guidelines to the few runtime contracts the outer schema cannot express (`null` failures, explicit `return`, `label` / `tier` / structured output);
5. shortens the tool and parameter descriptions without removing fields or capability; and
6. adds regression budgets for the rendered system-prompt contribution and serialized tool definition.

A schema-only experiment across four parent models suggests most current guidance is unnecessary. A compact version retained valid workflow selection, parsing, fan-out, synthesis, and inline execution. My current measurements indicate the system-prompt contribution can fall from **20,485 bytes to about 610 bytes**, while the provider-visible definition can fall from **2,204 bytes to about 1,467 bytes**.

[...]

PR #66, opened 2026-07-14
fix: remove model catalogue from workflow prompt

Whamp:
## Summary

- remove the live available-model catalogue from `promptGuidelines`
- keep normal routing on `small` / `medium` / `big`, with `opts.model` only for a user-named exact model
- add regression budgets around the accepted rendered prompt and provider-visible tool definition

The limits were established after the focused prompt form was complete: **6,500 UTF-8 bytes** for the isolated rendered workflow prompt contribution and **2,204 UTF-8 bytes** for `JSON.stringify({ name, description, parameters })`. They are exact observed ratchets, not the more aggressive research targets. The prompt baseline deliberately uses an empty `agentType` registry because removing that catalogue is deferred.

The provider-visible definition is unchanged in this PR; its test records the accepted current form so future growth is deliberate.

## Validation

- `npm test` — 825 passing
- Pi SDK integration test renders the real workflow prompt and reads the wrapped tool from `session.agent.state.tools`
- configured model IDs are checked before and after registry replacement and never enter `promptGuidelines`

This changes parent-model guidance only; runtime tier resolution and subagent model execution are unchanged, so no real-provider routing run was required.

Part of #65.

### DIFF
--- a/src/workflow-manager.ts
+++ b/src/workflow-manager.ts
@@ -165,10 +165,8 @@ export class WorkflowManager extends EventEmitter {
   }
 
   /**
-   * The host session's model registry, when set. Read lazily (e.g. by the
-   * workflow tool's model routing guideline) since `setModelRegistry` is called
-   * from `session_start`, which runs after the tool is created — a snapshot
-   * taken at tool-creation time would miss it.
+   * Expose the host session's model registry to integrations sharing this
+   * manager. Workflow execution reads the same registry internally.
    */
   getModelRegistry(): ModelRegistry | undefined {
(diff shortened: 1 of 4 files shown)


QuintinShaw:
Thanks @Whamp — this is exactly the incremental first step we discussed, cleanly scoped. I reviewed the diff and re-ran the suite locally (825 passing): the catalogue is gone from the always-on guidelines, tier routing and the exact-model path are untouched, and the byte-budget ratchets are a nice touch to keep the win from eroding over time. Merging now — really appreciate the careful research behind this.

PR #73, opened 2026-07-15
feat: show cached token reads separately across workflow displays

vssnake:
## What

The workflow token displays show a single scalar per agent — `input + output + cacheRead + cacheWrite` — which mixes real spend with cheap cache reads. For a ~95%-cached agent that reads as huge (e.g. "3.0M tok" when only ~100k was billed at full price); for a non-caching provider the same number is 100% real spend. Same unit, opposite meaning.

This splits it into `N tok · M cached` across every surface: live task panel, `/workflows` navigator (runs list, phase header, agent rows), status/print view, the `/workflows` text commands, the tool result, and the background-run delivery line.

**Before**
```
[1] ✓ paper-generation   3.0M tok    · claude-opus-4.8
[2] ✓ numeric-auditor    6.3M tok    · glm-5-2
```

**After**
```
[1] ✓ paper-generation   100K tok · 2.9M cached   · claude-opus-4.8
[2] ✓ numeric-auditor    6.3M tok                 · glm-5-2
```

`tok` = input + output + cacheWrite (cache writes are first-time ingestion billed at full or premium price, so they count as real spend); `cached` = cacheRead (cheap reuse), omitted when zero.

## What changed since #57 (and why it was closed)

#57 was closed because the split rendered inconsistently across setups: providers that don't report per-agent usage fell back to the old scalar, so the same workflow looked like two different features depending on the provider. This revision makes the degraded path first-class:

1. **The degraded mode is visually identical to the base format.** The wording changed from `N fresh · M cache` to `N tok · M cached`: an agent without cache data renders `N tok` — exactly the pre-existing scalar format. Cache data enriches the line when available and changes nothing when not.
2. **One fallback rule, one place.** `tokenFigures()` reconciles the two token sources the pipeline already had (the provider-reported breakdown, and the scalar estimate that keeps accruing when the provider reports nothing): fresh is the reported spend, but never less than what the estimate can account for after removing cache reads. Estimate-only providers, cost-only providers, and mixed runs all keep the count the display showed before the split.
3. **Zero never lies.** `usageFromStats()` skips `onUsage` for all-zero session stats, and `fmtTokenSegment()` centrally omits the token segment when nothing is known (e.g. a journal-replayed resume) — no surface can render a false "0 tok".
4. **The split survives persistence** — per-agent usage is carried through `persistedToSnapshot`, so cross-session runs render the same as live ones.

## Pre-existing display quirks this also fixes

Three small fixes ride along because the shared helpers make them free and they serve the same goal (truthful token accounting). Calling them out explicitly since they predate this PR:

- **Persisted runs never showed per-agent tokens** — `persistedToSnapshot` didn't copy them (this one is required for the split to work cross-session at all).
- **The `/workflows` runs list showed no token count for live runs** — the run-level aggregate only lands at completion; the list now uses whichever source (run aggregate vs per-agent figures) accounts for more tokens, so it agrees with the phase view.
- **A real but tiny cost could render as `$0.0000`** (`toFixed(4)`); `fmtCost` floors it at `<$0.0001`, and an all-zero usage aggregate no longer renders a literal `0 tokens`.

If you'd rather keep the PR minimal, I'm happy to split these into a follow-up.

## How

- Per-agent `AgentUsage` is read from `session.getSessionStats()` in a `finally` (fires on the error path too) and carried through `onAgentEnd` into `WorkflowAgentSnapshot.tokenUsage`.
- All surfaces render through shared helpers: `tokenFigures` (fallback rule) → `fmtTokenSegment` (blank-on-zero) → `fmtTokenCount` (format), plus `fmtCost` (unified cost format: 2 decimals ≥1¢, 4 below, `<$0.0001` floor).
- Backward compatible: snapshots without `tokenUsage` (including previously persisted runs) render the existing scalar count.

`feat:` because it adds public surface (`WorkflowAgentSnapshot.tokenUsage`, the `onAgentEnd` event's `tokenUsage`, and the display helpers). Display-only — nothing changes about what is sent to models or how much is spent. Two deliberate display-behavior changes ride along: the delivery line now includes the run cost, and cost formatting is unified across surfaces.

## Verification

`npm test` green on top of v2.13.1 (biome + tsc + 844 unit tests). The change went through two adversarial multi-agent review rounds; every finding is covered by a dedicated regression test: persisted-run split, estimate-only header/delivery, cost-only agents, mixed runs, all-zero aggregates, cache-write-heavy agents, live-run list aggregation, and the `tokenFigures`/`fmtCost` edge cases.

End-to-end on a real Pi install (dev-linked package) with a deliberately asymmetric setup — a vLLM gateway provider that reports usage but no caching, plus github-copilot/claude-opus-4.8 which caches heavily. A 2-agent demo run (screenshots below) renders the non-caching agent as a plain `N tok` (no dangling cache segment) next to a caching multi-turn agent with its `· N cached` split, and the runs-list row shows the aggregated split with cost. A separate 9-agent production run renders the header aggregate across six same-phase vLLM agents (plain counts) and cached Opus agents.

## Screenshots

two-pane-navigation
detail-worflow

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ return await agent('Synthesize and double-check these findings:\n' + findings.jo
 - **Journaled resume** — an interrupted run replays finished agents from a journal (no re-run, no tokens) and runs only what's left or what you changed.
 - **Git worktree isolation** — `isolation: "worktree"` gives an agent its own branch, so parallel agents can edit the same files without clobbering each other.
 - **Real token & cost accounting** — read from each subagent's session, not estimated. Runs have no default token cap; `tokenBudget`, phase budgets, and `budget` let you add explicit gates when you want them.
-- **Background by default** — the turn ends right away, a live "Workflows running" panel tracks runs, and each result is delivered back so the conversation auto-continues when it finishes. The panel is compact by default; `/workflows-progress detailed` expands it inline to per-phase/per-agent rows with tokens, cost, and a live tok/s rate (so a stalled agent shows as 0 tok/s) — no need to open `/workflows`.
+- **Background by default** — the turn ends right away, a live "Workflows running" panel tracks runs, and each result is delivered back so the conversation auto-continues when it finishes. The panel is compact by default; `/workflows-progress detailed` expands it inline to per-phase/per-agent rows with a fresh/cache token split, cost, and a live tok/s rate (so a stalled agent shows as 0 tok/s) — no need to open `/workflows`.
 - **Interactive `/workflows` TUI** — drill runs → phases → agents → detail; inspect per-agent failures and compact subagent history; pause, stop, restart, and save runs from the keyboard.
 - **Quality patterns built in** — `verify()`, `judgePanel()`, `loopUntilDry()`, and `completenessCheck()` for adversarial review, best-of-N, and exhaustive discovery.
 - **Ultracode** — `/ultracode` is a standing opt-in that auto-arms an exhaustive multi-agent workflow for every substantive message, the way Claude Code's ultracode does. `/effort high` is the lighter tier.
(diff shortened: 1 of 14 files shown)


QuintinShaw:
Thanks @vssnake — clean resubmission of #57 with the cross-provider degraded-mode cases now solid. Re-verified locally: the fresh-vs-cached split is consistent across every display, cacheWrite correctly counts as real spend (not cheap reuse), non-caching providers render cleanly with no dangling cache segment and no false '0 tok', and 844 tests pass. Nice self-cleanup commit too. Merging.

PR #74, opened 2026-07-15
feat: make workflow runs observable and resumable

SSS135:
## Summary

This PR hardens workflow triggering, lifecycle control, observability, persistence, and resume behavior as one runtime-reliability change:

- disable forced keyword triggering by default; retain explicit opt-in with identifier-bounded matching and preserve slash commands
- add the LLM-facing `workflow_control` tool for list/status/set_concurrency/pause/resume/stop/restart/remove
- persist requested/effective concurrency and all execution controls; concurrency defaults to `16`, accepts any positive integer without a plugin hard cap, and can be resized while a run is live or paused
- track stable agent execution identities, queued/running/paused/terminal states, active labels, live usage estimates, and exact message-boundary usage
- add versioned run-state migration, lease/race protection, replay-safe aggregate accounting, and restart/resume semantics
- persist child Pi sessions by default in a private per-project workflow directory outside Pi's normal `/resume` picker, and reopen paused agents at the last durable message/tool-result boundary; `persistAgentSessions: false` remains an explicit opt-out
- preserve isolated worktrees while paused, reuse them on resume, and clean them on completion/stop
- distinguish finished, paused-mid-run, and never-started agents throughout the panel, navigator, and control output

## Why forced keyword triggering is disabled by default

Previously, merely typing `workflow` or `workflows` in an ordinary message could rewrite the prompt and force workflow mode. That made it difficult to discuss the feature itself, mention documentation or paths containing the word, or use the word naturally without unexpectedly launching orchestration. Hidden prompt mutation based on one keyword is poor UX and produces false positives.

The lexical forcing hack is also unnecessary: current models can select the advertised `workflow` tool from its schema and system guidance when orchestration is appropriate. Explicit `/workflows run ...`, direct tool selection, saved workflow commands, and slash-command controls remain available. Users who still want keyword activation can opt in with `/workflows-trigger on`; when enabled, matching is Unicode/identifier-bounded rather than a raw substring check.

## Resume semantics

- Completed journal entries replay without rerunning or double charging.
- Started-but-incomplete agents resume from their saved Pi session and preserved worktree.
- Never-started agents remain queued and start fresh.
- Missing/corrupt session files fall back per-agent to a fresh restart.
- Exact mid-token/mid-tool continuation is not possible; the continuation prompt tells agents to inspect existing state and avoid repeating completed side effects.
- Crash recovery pauses stale runs instead of auto-resuming them.
- Resume continues the same run; restart creates a new run from the beginning.

## Default behavior changes

[...]

PR #79, opened 2026-07-16
fix: bound workflow keyword matching

SSS135:
## Summary

Adds Unicode-aware identifier boundaries for workflow keyword matching while preserving the existing default-on auto-trigger behavior.

- matches the built-in and custom trigger words at identifier boundaries
- handles Unicode identifiers consistently
- treats POSIX and Windows path components as bounded contexts
- keeps interactive keyword triggering enabled by default

## Windows test portability

Full Windows validation identified three test-harness portability issues on current `main`: home-directory isolation across `HOME` and `USERPROFILE`, platform-aware expected paths with `node:path.join`, and file-URL conversion for absolute dynamic-import paths. The corrections are confined to test setup and path handling and preserve the existing assertions.

## Validation

- `npm test`: **899/899 passed**
- focused workflow-editor suite: **138/138 passed**
- Biome: passed
- TypeScript build: passed
- `git diff --check`: passed
- independent shortcut, quality, and spec reviews completed with no scoped implementation findings

Split from and supersedes the keyword-matching portion of #74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Run a workflow to audit every route under src/routes/ for missing auth checks.
 
 Pi writes and starts the workflow in the background. A live panel tracks progress while you keep working, and the final result is delivered back into the conversation automatically.
 
-Use the word **workflow** or **workflows** in a message to force workflow mode, or run `/workflows run <prompt>` explicitly. If the default keyword is too broad, change it with `/workflows-trigger set pi-workflow` or disable it with `/workflows-trigger off`.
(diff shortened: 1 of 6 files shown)


QuintinShaw:
Thanks @SSS135 — clean first split, and I appreciate you rebasing onto 2.14.0 already. Reviewed closely: the Unicode identifier-boundary regex is correct and ReDoS-safe (fixed-length lookarounds + a literal match, no nested quantifiers), default-on keyword triggering is genuinely preserved (only a new test locks it in — the `?? true` default logic is untouched), and the Windows test-portability fixes are test-only. Tightening the default `workflow` word to identifier boundaries (so `workflowRunner` / `my-workflow-helper` no longer trigger) is a nice correctness bump on top. 899 tests green.

Merging. Looking forward to the rest as you mark them ready — `workflow_control` (#83) looks like the next clean standalone one; the two big resume PRs (#80/#81) will want a rebase since #78/#84 just landed in that area.

PR #83, opened 2026-07-16
feat: add workflow_control and run observability

SSS135:
## Summary

Adds a model-facing `workflow_control` tool with structured run and agent observability, built on the lifecycle methods in `WorkflowManager`.

Supported actions:

- `list`
- `status`
- `pause`
- `resume`
- `stop`

Status reports the canonical run ID, workflow name, lifecycle state, current phase, agent counts, active labels, and the most complete token total available across live, persisted, and per-agent usage.

## Validation

- `npm test`: **908/908 passed** across 64 suites
- Biome: passed
- TypeScript build: passed
- `git diff --check`: passed
- independent shortcut, quality, and spec review completed; the token-aggregation finding was addressed

This focused PR is split from and supersedes the control and observability portion of #74.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,9 @@ return await agent(
 
 For an always-on exhaustive mode, use `/ultracode`; `/effort high` is the lighter standing option.
 
-## Commands
+## Commands and run control
+
+Pi can manage background runs directly with the `workflow_control` tool instead of asking you to type a command. It supports `list`, `status`, `pause`, `resume`, and `stop`; run-specific actions use the canonical run ID returned when the workflow starts. Status output includes the run state, current phase, agent counts, active labels, and recorded token total.
 
 | Command | Purpose |
 | --- | --- |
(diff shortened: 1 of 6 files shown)


QuintinShaw:
Thanks @SSS135 — clean split, and it composes exactly as hoped: `workflow_control` is a thin wrapper over the existing manager methods (`list`/`status`/`pause`/`resume`/`stop`), `resume` goes through the same lease-guarded path so it's race-safe against the auto-resume scheduler, no `PersistedRunState` changes, and scope stayed tight — the deferred default-flips aren't here, and you even reject `restart`/`remove` rather than half-implement them. Reviewed and verified 908 tests.

It doesn't touch `agent.ts`/`package.json`, so it's zero-overlap with the 0.80.8 `ModelRuntime` migration (#91) that just landed — merging as-is (main keeps the `>=0.80.8` peerDep). Thanks for taking the split on; looking forward to the rest as you mark them ready.

PR #91, opened 2026-07-17
fix: support pi >= 0.80.8 ModelRuntime API (removed AuthStorage/ModelRegistry.create)

RyanSandstone:
## What broke

pi 0.80.8 shipped breaking SDK changes ([CHANGELOG](https://github.com/earendil-works/pi/blob/main/CHANGELOG.md#0808---2026-07-16)):

- `AuthStorage` is no longer exported from `@earendil-works/pi-coding-agent` (only `readStoredCredential` remains).
- `ModelRegistry.create(auth, modelsPath)` is gone — `ModelRegistry` is now a synchronous facade constructed from an async-created `ModelRuntime`.
- `CreateAgentSessionOptions.modelRegistry` was replaced by `modelRuntime`.

Because pi loads extensions via jiti (missing named imports become `undefined` instead of a load error), the package still _loads_ on 0.80.8+, but degrades silently:

1. **Disk-fallback registry paths throw at runtime** — `AuthStorage.create(...)` is `undefined.create(...)` (`listAvailableModels()` with no registry, `WorkflowAgent.getRegistry()` with no shared registry).
2. **Subagents lose the host session's catalog** — `createAgentSession` silently ignores the removed `modelRegistry` option, so every subagent gets a default-built runtime. Models from extension-registered providers (e.g. an ollama extension) resolve at spec time but are unreachable at stream time.
3. **The usage-limit integration tests fail** — `registerFauxProvider()` registers into the compat api registry, but a `ModelRuntime` session with an un-overlaid builtin provider never consults that registry (`recomposeProvider` uses the builtin untouched), so the faux turns are never served and the tests die on real connection errors after ~14s of retries.

The repo's own `npm test` currently only stays green because `package-lock.json` pins `@earendil-works/pi-coding-agent@0.80.6`; a fresh `npm update` to latest breaks the build on `import { AuthStorage }`.

## The fix

- **`src/agent.ts`**
  - Disk fallback is now a lazily-created shared `ModelRuntime` (`ModelRuntime.create({authPath, modelsPath})`) wrapped in a `ModelRegistry` facade; the availability snapshot is warmed once so the facade's sync `getAvailable()` is populated. A rejected create is not cached (transient failures don't wedge the process). Sync callers (`listAvailableModels`) report `[]` while it initializes — same best-effort contract as before, and the workflow tool's lazy `promptGuidelines` re-reads pick up real specs on the next read.
  - `getRegistry()` is now async (registries wrap an async-created runtime); `run()` resolves the registry once and uses it for tier diagnostics, model resolution, and the session.
  - `run()` passes the registry's underlying `ModelRuntime` to `createAgentSession` (replacing the removed `modelRegistry` option), so subagents share the host session's exact catalog and auth again. `ModelRegistry` doesn't expose its runtime publicly, so this reaches the private `runtime` field via an exported `runtimeOf()` helper that warns once and degrades to the previous (default-runtime) behavior if pi ever renames the field.
- **`tests/usage-limit-integration.test.ts`** — rewritten for the `ModelRuntime` dispatch path: the faux core (`createFauxCore`) is registered as an extension provider (`registerProvider` with `api` + `streamSimple`) on an explicit test-scoped `ModelRuntime` handed to the session. Faux state is closure-scoped, so no global registry or module-instance concerns; tests now run in milliseconds instead of retrying real connections.
- **`tests/model-runtime-routing.test.ts`** (new) — pins the private-field contract against the real installed pi, plus an end-to-end test where a subagent reaches an extension-registered provider **only** via `runtimeOf()` (no `session.modelRuntime` override). If a future pi renames the internals, these fail loudly instead of tsc/mock tests staying green while routing silently degrades.
- **`tests/agent.test.ts`** — updated to the async `getRegistry()` contract.
- **`package.json` / `package-lock.json`** — peer dependency bumped to `@earendil-works/pi-coding-agent >= 0.80.8` (`ModelRuntime` doesn't exist earlier) and the lockfile refreshed to 0.80.10. If you'd rather keep supporting 0.80.6/0.80.7, `runtimeOf`/the fallback could feature-detect `ModelRuntime` vs `AuthStorage` — happy to amend, but given pi's release cadence a minimum-version bump seemed cleaner.

## Verification

- `npm test` (biome + tsc build + unit tests) green against pi **0.80.10**: 904/904.
- Per CONTRIBUTING, verified end-to-end against real pi 0.80.10 subagent sessions (real `createAgentSession` → real provider): plain `agent()`, tier routing (`small`/`big` resolving to different real models), and `schema` structured output all work; results and token accounting come back correctly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -699,9 +699,9 @@
       }
     },
     "node_modules/@earendil-works/pi-ai": {
-      "version": "0.80.6",
-      "resolved": "https://registry.npmjs.org/@earendil-works/pi-ai/-/pi-ai-0.80.6.tgz",
-      "integrity": "sha512-7xfLk8sANBp+bpPEbjoOZTbPxsa+++b1JXAoSJsNa3vbs9AHHEclmvg54XLQcxH+fuwaeti/g2jeIfJ+mVYLpA==",
+      "version": "0.80.10",
+      "resolved": "https://registry.npmjs.org/@earendil-works/pi-ai/-/pi-ai-0.80.10.tgz",
+      "integrity": "sha512-Moe/H8c87yacDGK9dPbWphZNjVsrb3nTrIHycOQJAkFEnY9PYxOOd74+ny44kATfPU9Dm7aTHefar3pZF+UKUA==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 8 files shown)


RyanSandstone:
Pushed one more commit covering a first-use regression the same 0.80.8 change causes in `/workflows-models`: the command built its default tier config via bare `listAvailableModels()`, which now initializes its disk fallback asynchronously and reports `[]` on the first call — so a first-ever open ranked all tiers from an empty model list. The handler now passes `ctx.modelRegistry` at both call sites (initial default build + "Reset to defaults"), with a handler-level regression test that drives the real command against a stub registry (fails without the fix, passes with it). `npm test`: 905/905.

QuintinShaw:
Thanks @RyanSandstone — thorough and well-timed fix. I reviewed it closely: the migration covers every affected call site (the `listAvailableModels`/`getRegistry` disk fallbacks, the `createAgentSession` → `modelRuntime` switch, and the `workflows-models-command` registry plumbing), and since it's rebased on current main nothing recent is missed. Making `getRegistry` async is threaded through correctly — `run()` awaits it once and reuses the result (so the tier-fallback warning path doesn't get handed a Promise), and the tests move from `doesNotThrow` to `doesNotReject` to match.

Nice touch that the new tests exercise the real SDK classes end-to-end to tripwire a future rename of the private `runtime` field, and that you sidestep the removed `AuthStorage` entirely by passing `authPath` into `ModelRuntime.create`. Bumping the peerDep floor to `>=0.80.8` is the right call for a hard compat break. 905 tests green.

Merging — this ships in a patch release right away, since 0.80.8+ users are silently degrading today.

PR #92, opened 2026-07-17
feat: improve workflow detail pager

yycholla:
## Summary

- show a compact agent-detail summary by default: completed agents show their final result, while active/failed agents show their prompt and two latest history events
- retain full agent results and expose an explicit full pager with Markdown/code highlighting, line/page/end navigation, live tail mode, and bounded short-terminal rendering
- keep long run lists focused on the selected item, coalesce relevant `agentHistory` refreshes, cache expensive rendering/run reads, and document the new navigator controls

## Details

The full pager can be opened from any agent detail with `Enter`/`Right`. It supports:

- `j`/`k` or arrow keys for line scrolling
- `PgUp`/`PgDn` (plus common control-key aliases) for page scrolling
- `g`/`G` or `Home`/`End` for boundary jumps
- `t` for live tail/follow mode
- `Enter` or `Esc` to return to the compact summary

Markdown results use Pi's Markdown renderer. Structured results are rendered as highlighted JSON, saved workflow scripts as JavaScript, raw `read` history inherits syntax highlighting from the requested file path, and `write` calls render their target path and source directly instead of dumping a raw JSON argument envelope.

To keep large run histories responsive, each navigator frame shares one persisted snapshot, persistence briefly caches repeated list reads, Markdown/highlighting output is cached per overlay, and history redraws are limited to the selected agent. Terminal runs also store each full agent result only once instead of duplicating it in both the agent snapshot and resume journal.

## Testing

- `npm test` — 942 tests passed
- added coverage for compact completed/running/error summaries, de-duplicated full-result persistence, per-frame/list caching, pager toggling and back behavior, line/page/end key mappings, tail following, raw-read and write-source syntax highlighting, short viewports, and long run lists
- manually exercised with real Pi subagent workflows that performed multiple source reads and wrote/read substantial temporary TypeScript fixtures

## Compatibility

- no new configuration or workflow tool parameters
- `WorkflowAgentSnapshot.result` is optional
- existing saved runs continue to render through `resultPreview` when no full result is available


QuintinShaw (changes requested):
Really nice work — the pager, the syntax highlighting, and the caching (bounded LRU, content-keyed) are all clean, and the 923-test suite is green.

One blocking issue before merge, in the result-dedup change. `persistRun` now strips `agents[].result` for any non-terminal run (`keepsResumeJournal` is true for running/paused/**failed**), leaving the full value only in the journal. But `persistedToSnapshot` — the cold-read path used whenever `getRun()` misses, i.e. after a pi restart or from another session — only reads `agents[].result`/`resultPreview` and never consults `p.journal`. So once you restart pi and open `/workflows` on a previously failed/paused/interrupted run, its completed agents show only the truncated preview, even though the full result is still on disk in the journal. It's not a crash and the bytes aren't lost, but it's a durability/UX regression vs `main` (which persisted the full result into `agents[]`) and it undercuts exactly the after-the-fact inspection the pager is for. Same-session live views are unaffected.

Suggested fix: in `persistedToSnapshot`, when a `done` agent has no `result`, reconstruct it from `p.journal` by positional call index (journal `.index` is the call index — the same match `resume()` already relies on). A regression test that persists a non-terminal run, re-reads it via a fresh manager, and asserts the pager sees the full result would lock it in — the current tests only exercise completed/live paths.

Two minor, optional: Right on an already-open pager toggles it closed (the description frames Right as "open"); and the shared `historyRenderTimer` can drop a coalesced history update when you switch agents mid-window (self-heals on the next event). It also needs a quick rebase onto current `main` (disjoint files, should be clean).

Happy to take the journal-fallback fix myself and push to your branch if you'd prefer — your call.

yycholla:
@QuintinShaw Fix is in

- Right input now only opens the pager; Enter still toggles it.
- History redraw tracks the latest active agent, preventing dropped updates after switching agents.

yycholla:
Updated the branch to current `main` (`v3.3.0`) and resolved the conflicts.

Conflict resolution preserves the pager changes together with the newer navigator hardening/event gating. The cold-read journal fallback now uses namespaced call identities (with positional fallback for legacy files), so parent and nested `index: 0` results cannot be mixed. Added regression coverage for that case.

Validation:
- `npm test` — 1137 passed
- final `npm run build`
- final targeted manager/pager tests — 124 passed

The PR is conflict-free and ready for another review.

QuintinShaw:
Thanks for the careful conflict resolution — reusing the `runId:index` namespaced identity for the cold-read fallback instead of inventing a parallel scheme was exactly right, and the regression test for the parent/nested index-0 case genuinely bites. The `keepsResumeJournal` invariant keeping a single on-disk copy of results is a nice touch. Merging.

Issue #89, opened 2026-07-17
Pi behavior is weird when workflow execution is triggered via the editor component keyword.


rwese:
Hi, currently, when a workflow is triggered from the editor component, the Pi process becomes idle, and you have to use the `/workflow` command to inspect the running workflow.

When workflows are used without the trigger being enabled, the LLM calls the tool itself and displays the workflow progress inline. In my opinion, this provides a much better experience.

I would prefer the editor-triggered workflow behavior to be more consistent with the normal workflow tool usage. Ideally, the Pi process would remain active while the workflow is running, so the interaction continues and users can queue additional prompts.

So the solution is to disable the trigger, but loses the nice animation.

A solution would be to have the trigger word _enable_ the workflow tool, and prompts. Currently this extension adds a lot of system prompt tokens, all the time.

QuintinShaw:
Thanks @rwese — sharp observation, and it lines up with #88. You've put your finger on the core issue: a keyword-triggered run currently *rewrites your message and forces a background workflow*, so Pi goes idle and you have to `/workflows` to watch it — whereas when the LLM calls the tool itself you get the inline experience and can keep interacting.

Your suggestion — have the trigger word *enable* the workflow tool (and let the model call it inline) rather than force a background run — is the right direction, and it also eases the always-on prompt-token cost you mentioned (@Whamp is separately trimming that in #65). Together with #88's over-triggering/default-on feedback, I'm treating keyword-trigger behavior as a design track to reshape rather than patch. Thanks for laying it out so clearly.

Whamp:
Well enabling the tool part way through a session will break prompt cache and either get very expensive or make your subscription provider angry. I'm working on changes that teach the models it has the option to use either a background non blocking workflow or a blocking workflow. Part of the existing instructions in the tool description tell it to default to background workflows but don't tell it how which is one of the things I'm fixing in my PR. 

Quick update on that. I'm moving lots of tool instructions to a skill to teach the model the available contracts and am running auto research to make it as concise as possible while still teaching a dumb model to execute it and pass all my workflow writing tests. Which hopefully means smarter models will have no problem.

QuintinShaw:
[...]

PR #100, opened 2026-07-18
feat: make workflow capabilities self-documenting

### LINKED ISSUE #65: Reduce always-on workflow prompt size (model catalogue adds ~5k tokens)

Whamp:
## Summary

- add one executable capability contract for workflow globals, tool inputs, option shapes, lifecycle constraints, static references, and dynamic catalogue ownership;
- assemble project-owned workflow globals through that contract while preserving the existing runtime closures and VM behavior;
- ship a package-discovered `workflow-authoring` skill with focused references and 11 runtime-tested examples;
- generate capability publications and context measurements, then enforce version, package, link, coverage, runtime-alignment, and freshness checks during normal tests and publishing;
- keep provider-backed comprehension optional while adding model-free replay, capability-coverage protection, and delivery-choice evidence.

Closes #65.

Implements the capability-contract design from [Whamp/pi-dynamic-workflows#22](https://github.com/Whamp/pi-dynamic-workflows/issues/22). Adds evidence for the background-versus-inline decision discussed in #89 without changing #93's trigger authorization behavior.

## Context impact

Compared with `origin/main` at `818fdd8`:

| Surface | Main | This PR | Change |
| --- | ---: | ---: | ---: |
| Permanent workflow prompt | 766 bytes | 742 bytes | -24 |
| Provider-visible workflow tool definition | 9,558 bytes | 3,802 bytes | -5,756 |
| Skill discovery | 0 bytes | 338 bytes | +338 |
| Ordinary workflow-owned always-on total | 10,324 bytes | 4,882 bytes | **-5,442 (-52.7%)** |

Detailed authoring guidance now loads on demand. In the comparative nine-model coverage sample, the optimized skill loaded a median 2,564 tokens versus 8,096 for the pre-trim skill, a 68.3% reduction.

See [`docs/workflow-prompt-guidance-rationale.md`](docs/workflow-prompt-guidance-rationale.md) for the reviewed record of prompt insertions, removals, and compactions. See [`docs/workflow-authoring-evidence.md`](docs/workflow-authoring-evidence.md) for methodology, model matrix, per-scenario results, and reproduction commands.

## Capability and quality evidence

- The capability contract inventories the 18 curated project-owned globals and all current workflow-tool inputs, including `resumeFromRunId`.
- Runtime assembly rejects missing declared implementations, reports and ignores undeclared ones, and preserves implementation identity.
- Live model and `agentType` values remain outside static prompts and generated references.
- The installed skill contains focused write/edit/review/debug routes, exact generated facts, six orchestration patterns, four focused recipes, and a dedicated validated-`gate()` recipe.
- Every stable authoring surface is either covered behaviorally or protected as guidance-frozen.
- Optional comprehension prompts do not name the skill; generated workflows run through the real parser/runtime with deterministic fake agents.

Post-tuning provider-backed validation:

- pre-trim skill: **73/81**;
- optimized skill: **74/81**;
- generate-and-filter: **27/27** optimized versus **26/27** pre-trim;
- all seven configurations other than Qwen and Spark: **63/63** optimized;
- protected GPT-5.4 Mini/Luna coverage gate: **18/18**;
- protected GPT-5.4 Mini/Luna six-scenario core gate: **36/36**.

This is not an independent holdout: the same scenario families informed the guidance changes. An earlier corrected-harness sample scored pre-trim at 75/81 and optimized at 71/81, so provider scores remain sampled, non-blocking evidence. The final candidate did not repeat the earlier nine-model, 162-case core panel.

The delivery-choice harness has deterministic tests and an optional provider CLI; this PR reports no provider-run delivery-choice result. Provider calls are not part of normal tests, release checks, or publishing.

## Compatibility

- preserves #93's single opt-in `promptGuidelines` gate and default background delivery;
- preserves the existing truthy-object `gate().ok` compatibility path;
- preserves the existing regex determinism blocklist and VM substrate;
- keeps `console` available for saved workflows while teaching `log()`;
- changes no live model-route or `agentType` configuration format;
- adds no migration state, feature flag, or dual-running path.

## Verification

- `npm run check`
- `npm run build`
- `npm test` — **1,027/1,027 passed**
- `npm run docs:check`
- `npm run context:check`
- `npm run guidance:check`
- `npm run release:verify` — **0 warnings**
- `npm pack --dry-run` — 166 files; all 27 workflow-authoring skill files included
- `git diff --check`
- CodeGraph changed-file cycle and boundary checks

The PR is one atomic commit based on current `origin/main`. Raw `.auto` experiment data is intentionally excluded from both the commit and package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ dist/
 scripts/verify-npm-install.sh
 __pycache__/
 .context/
+.auto/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+# Repository guidance
+
+## Workflow documentation
+
+Before changing the workflow runtime, tool API, capability contract, or `workflow-authoring` skill, read [Protected workflow-authoring guidance](CONTRIBUTING.md#protected-workflow-authoring-guidance).
+
+- Keep stable capability facts in the executable capability contract and generated documentation.
+- Keep detailed authoring guidance in the on-demand skill, not the always-on prompt.
+- Do not copy live model or agent-type catalogues into static guidance.
+- Run `npm run context:check` with the other checks listed in the contributor guide.
+- If a protected file changes, review it before running `npm run guidance:accept -- <path>`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,6 +29,8 @@ A throwaway harness for this should live in the repo root (not `/tmp`, whose sym
 
 ## Protected workflow-authoring guidance
 
+The goal is to keep workflow guidance accurate as the runtime changes without bloating the context sent to every model. Stable capability facts come from the executable capability contract and generated documentation. Detailed authoring guidance lives in the on-demand `workflow-authoring` skill instead of the permanent prompt. Context checks keep these surfaces from growing unnoticed.
+
 Some files under `skills/workflow-authoring/` contain mixed or partially behavior-covered guidance. Their full-file SHA-256 hashes in `WORKFLOW_AUTHORING_FROZEN_FILES` (`src/workflow-authoring-coverage.ts`) are explicit review checkpoints, not proof that the wording is correct.
 
 If `PROTECTED_GUIDANCE_DRIFT` reports an accidental change, revert it. For housekeeping such as a typo, link, formatting, or version update, deterministic checks and review are enough. For a semantic guidance change, inspect the affected coverage manifest entry, update relevant behavioral tests, and review provider evidence when needed. Required anchors and required text in the manifest may also need deliberate updates.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ Thanks for contributing to pi-dynamic-workflows. This project values small, well
 
 ```bash
 npm install
-npm test     # biome check + tsc build + unit tests — must pass
+npm test     # Biome, TypeScript, unit tests, and release checks — must pass
 ```
 
 `npm test` runs exactly what CI runs. If it's green locally it should be green in CI. CI runs on every PR to `main`; for fork PRs a maintainer approves the first run.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,6 +27,21 @@ Fake-agent unit tests are necessary but not sufficient. Any change to how agents
 
 A throwaway harness for this should live in the repo root (not `/tmp`, whose symlink breaks relative imports), import from `./src`, and be deleted before commit — don't commit harnesses.
 
+## Protected workflow-authoring guidance
+
+Some files under `skills/workflow-authoring/` contain mixed or behaviorally uncovered guidance. Their full-file SHA-256 hashes in `WORKFLOW_AUTHORING_FROZEN_FILES` (`src/workflow-authoring-coverage.ts`) are deliberate review barriers, not generated output.
+
+If `PROTECTED_GUIDANCE_DRIFT` reports an accidental change, revert it. For an intentional change, inspect the affected coverage manifest entry and its relevant behavioral tests and provider evidence before accepting it. Required anchors and required text in the manifest may also need deliberate updates. Only after that review should you recompute the exact SHA-256 and manually update the matching `sha256` in `WORKFLOW_AUTHORING_FROZEN_FILES`; never change a hash only to make the gate green.
+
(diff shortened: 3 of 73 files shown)


QuintinShaw (commented):
Seriously impressive piece of work, @Whamp. Reviewed and tested it end to end — the runtime capability contract, #93 preservation, the skill + examples, and the new release gates, plus a real-pi smoke on 0.80.10 — and it's in excellent shape: #93's arm/authorize path is untouched, the tool schema stays a single top-level object (no repeat of the recent provider-400), the runtime globals are behavior-preserved, 1027 tests and `release:verify` are green, and the roughly 9.5KB → 3.8KB context win is real and measured. The evidence discipline — generated docs checked for drift, examples that actually run, provider calls kept out of CI — is exactly right. Thank you for the depth here.

One small fast-follow, non-blocking: there's no refresh path for the frozen-guidance hashes — editing `SKILL.md` (or a frozen reference/example) trips a hard `PROTECTED_GUIDANCE_DRIFT` with no `*:generate` to recompute the hardcoded SHA-256s, and the error doesn't hint how to fix it. Is that manual step deliberate — a conscious "you're touching protected guidance" gate, in which case a line in CONTRIBUTING would cover it — or would you want a small `guidance:freeze` script to match your other `*:generate` surfaces? Happy to send either your way.

Otherwise this is good to go — give me the nod and I'll merge.

Whamp:
Ok thank you and great feedback let me see what I can do to clean that up.

Whamp:
@QuintinShaw I simplified this check and documented why it exists.

The goal is to keep workflow guidance accurate as the runtime changes without putting the full authoring guide in every model prompt.

A protected file still fails the release check until someone reviews and accepts the change:

```bash
npm run guidance:accept -- <file>
```

The command updates only that file and shows the old and new hashes. `CONTRIBUTING.md` explains when tests are enough and when model testing is also needed. A new root `AGENTS.md` sends coding agents to these rules before they change the workflow runtime, tool API, capability contract, or authoring skill.

All 1,029 tests pass, and CI is green. Ready to merge.

QuintinShaw:
The guidance-acceptance flow is exactly right — thanks for turning it around so cleanly. Merging now.

PR #122, opened 2026-07-20
fix: preserve active workflows across Pi reload

yycholla:
## Summary

- keep the live `WorkflowManager` and session `/effort` state across Pi `/reload`
- refresh generation-bound dependencies, including the `ExtensionAPI`, delivery settings loader, toolsets, and manager defaults
- guard the handoff with the package version, pausing live runs for journal recovery after any extension update
- limit handoff to in-process reloads so process restarts continue using durable journal recovery

## Implementation

Pi creates a fresh extension generation during `/reload`, but active workflows still depend on the old manager's in-memory promises, abort controllers, leases, and event streams. This change stages that live runtime in a process-local, cwd-keyed handoff during `session_shutdown` when the reason is `reload`, then claims and reconfigures it from the fresh extension factory.

The handoff is single-use and expires after 30 seconds. Reloads within the same package version retain the exact live manager. If the installed package version changes, running workflows are paused and a fresh manager is loaded, avoiding mixed-version execution while leaving those runs safely resumable. Non-reload shutdowns discard the handoff. Existing durable restart behavior is unchanged: interrupted workflows recover from the journal as paused.

## Test plan

- [x] `npm test` (1,128 tests passed; release verification passed)
- [x] reload handoff is single-use, cwd-isolated, identity-guarded, and package-version guarded
- [x] the extension reuses the live manager across reload generations
- [x] manager defaults and result-delivery dependencies refresh after reload
- [x] interactive smoke test: start a background workflow, run `/reload`, then verify progress, controls, completion delivery, and `/effort`

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,8 @@ Subagents are in-memory by default. Set `persistAgentSessions: true` to retain f
 
 Completed background runs persist their full result in the project run JSON. The conversation delivery includes a pointer to that file when the visible summary is shortened.
 
+Pi's `/reload` keeps the live workflow manager in-process when the installed extension version has not changed. Active background runs therefore continue streaming progress, remain controllable, and deliver their result through the freshly loaded extension; session-local `/effort` also survives the reload. If the package version changes, active runs are paused and a fresh manager loads, leaving them safely resumable through the journal instead of mixing extension versions. This handoff applies only to `/reload`. A process restart still uses the same durable journal path, recovering an interrupted running workflow as paused so it can be resumed safely.
+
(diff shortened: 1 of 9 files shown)


QuintinShaw:
Thanks — this is carefully done. Keeping the same manager instance and refreshing only the generation-bound dependencies is the right shape, and the version-mismatch path pausing through the normal journal flow means nothing is stranded. Two small follow-ups you could take or leave: the 30s TTL expiry currently discards a still-running runtime without pausing it (in-flight subagents run to completion unobserved), and the "non-reload shutdown discards the handoff" guard has no test pinning it. Neither blocks merge. Merging.

PR #123, opened 2026-07-21
fix: declare args as an explicit object schema, not typeless Type.Any()

bruschill:
## Bug

The `workflow` tool's `args` parameter — used to pass arguments to the 5 named built-in patterns (e.g. `{ scope, checks }` for `codebase-audit`, `{ question }` for `deep-research`) — is declared with `Type.Any()`:

```ts
args: Type.Optional(
  Type.Any({ description: "Optional JSON value exposed to the workflow script as global `args`." }),
),
```

`Type.Any()` compiles to a JSON Schema with **no `"type"` keyword at all** — just `{ "description": "..." }`. Confirmed directly:

```js
const { Type } = require("@sinclair/typebox");
JSON.stringify(Type.Any({ description: "x" }));
// => {"description":"x"}
```

At least one MCP/tool-calling bridge does not treat a typeless property as "accept any JSON value" — it coerces/flattens the value before the tool handler ever sees it. The object a caller sends for `args` never reaches `resolveWorkflowInvocation()` intact:

```ts
function asRecord(args: unknown): Record<string, unknown> {
  return args && typeof args === "object" ? (args as Record<string, unknown>) : {};
}
```

`typeof args === "object"` fails, `asRecord` silently returns `{}`, and every built-in pattern's required-field validation throws — e.g.:

```
Built-in workflow "codebase-audit" requires args.scope to be a non-empty string.
```

...regardless of what was actually sent. Reproduced identically on `deep-research` with a single trivial `{"question": "test question"}` arg, so it's not pattern-specific — it makes `name`-based invocation of **all 5 built-in patterns** unusable end-to-end on that bridge. Only the `script` path (a plain typed `string` parameter, unaffected by this) still worked, which is exactly the workaround I used before digging into root cause.

This surfaced right after #117 made the 5 built-in patterns reachable via `name` + `args` from the tool itself (previously only via slash commands, whose args arrive as a plain string and go through a different parsing path — `tokenizeArgs` — untouched by this).

## Fix

Every built-in pattern's `args` is a JSON object at the top level (never a bare string/array/number), so this declares it as an explicit, minimal object schema instead of leaving it typeless:

```ts
Type.Unsafe<Record<string, unknown>>({
  type: "object",
  description: "Optional JSON value exposed to the workflow script as global `args`.",
}),
```

`Type.Unsafe` avoids `Type.Object({}, { additionalProperties: true })`'s extra `"properties":{}` boilerplate (JSON Schema already permits additional properties by default), keeping the byte cost to the provider-visible tool definition as small as possible: **+16 bytes** (4,267 → 4,283).

## Other changes required to keep the repo's own guardrails green

- `WorkflowToolInput.args` widened from `unknown` to `Record<string, unknown>` to match the now-typed schema (TS compile error otherwise).
- `TOOL_DEFINITION_BUDGET_BYTES` in `tests/workflow-prompt-budget.test.ts` bumped 4,267 → 4,283 (+16), documented inline in the same ratchet-comment style as the existing #105/#117 entries.
- Regenerated `docs/workflow-context-surfaces.json` (byte-size snapshot) and `docs/workflow-guidance-baseline.json` (content-hash snapshot) via their existing `npm run context:generate` / `npm run guidance:generate` scripts — both are generated artifacts tracked for drift detection.
- Added a regression test (`tests/workflow-tool.test.ts`) asserting `tool.parameters.properties.args.type === "object"`, so this can't silently regress back to a typeless schema.

## Verification

- `npm run build` — clean, no type errors.
- `npm run check` — clean (pre-existing lint infos unrelated to this change, untouched files).
- `npm test` (build + docs:check + context:check + full unit suite + release:verify) — **1124/1124 passing**, release gate green.

### DIFF
--- a/docs/workflow-context-surfaces.json
+++ b/docs/workflow-context-surfaces.json
@@ -13,7 +13,7 @@
     },
     "providerVisibleWorkflowToolDefinition": {
       "serialization": "UTF-8 bytes of JSON.stringify({ name, description, parameters })",
-      "bytes": 4267
+      "bytes": 4283
     },
     "registeredSkillsDiscovery": {
       "serialization": "sum of UTF-8 bytes of normalized Pi skill XML (name + description + location) across every root in package.json's pi.skills",
@@ -33,7 +33,7 @@
     },
     "ordinaryWorkflowOwnedAlwaysOn": {
       "serialization": "sum of permanent prompt, provider-visible tool definition, and every registered skill's discovery bytes",
-      "bytes": 5940
+      "bytes": 5956
     },
     "workflowAuthoringSkillCorpus": {
       "serialization": "sum of UTF-8 bytes for every file under skills/workflow-authoring",
--- a/docs/workflow-guidance-baseline.json
+++ b/docs/workflow-guidance-baseline.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "algorithm": "sha256",
   "surfaces": {
-    "compactGuidance": "aff3c746a9180ec9551030606d328008704234f62f3dade489a031fd4dbc8731",
+    "compactGuidance": "6db7b76e4e4017674cbad91c28c51132cef03121da20e1a23d3561f5652945cf",
     "detailedProse": "22fccb6ce35da667e7d57ed884d445e88b214f3c1d1890c3237f3c905c944da7"
   }
 }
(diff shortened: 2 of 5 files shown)


QuintinShaw:
Thanks — excellent root-cause work. Verifying that `Type.Any()` serializes with no `type` keyword at all, and picking `Type.Unsafe` to keep the tool-definition cost at +16 bytes, was exactly the right call. The regression test locking `args.type === "object"` is appreciated. One non-blocking note: the contract text in `workflow-capability-contract.ts` and the README still describe `args` as `unknown` — arguably still true at script runtime, so I'm leaving it as is for now. Merging.

PR #127, opened 2026-07-22
fix: make workflow token budgets opt-in

Whamp:
## Summary

- Make `tokenBudget` an opt-in soft spend gate and tell models never to infer it as a planning target.
- Align the workflow-authoring skill, lifecycle guidance, phased-budget example, and HIGH/ULTRA effort directives with that policy.
- Extend the #89 choice harness to verify omitted ordinary budgets and exact explicit caps across both `script` and named workflow invocations.
- Preserve the existing provider-visible definition ceiling: 4,276 bytes against the accepted 4,283-byte limit.

Addresses #89.

## Evidence

- `npm run build`
- `npm run docs:check`
- `npm run context:check`
- `npm run guidance:check`
- Focused changed-area tests — 160/160 before final review fixes; affected tests rerun after each fix
- `npm run delivery-choice -- --model openai-codex/gpt-5.6-sol:high` — 3/3 scenarios passed
- `npm test` — 1,149/1,150 locally. The only failure is the unchanged faux-provider object-schema test at `tests/agent.test.ts:335`, which also fails when run alone and is outside this diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,9 +310,10 @@ The comprehension harness is manual and never runs in normal CI, `npm test`, or 
 npm run comprehension -- --model provider/model                    # quick writing scenario
 npm run comprehension -- --model provider/model --suite full       # write, edit, review, and debug
 npm run comprehension -- --model provider/model --output runs/a.json
+npm run delivery-choice -- --model provider/model                  # timing and token-budget choices
 ```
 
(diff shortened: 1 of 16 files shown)


QuintinShaw:
Thanks — this is exactly the right fix for the #89 feedback, and fast. Making `tokenBudget` a pure opt-in gate and telling models it's not a planning target hits the root cause rather than papering over symptoms, and sweeping the skill, lifecycle doc, examples, and the HIGH/ULTRA effort directives in one pass keeps everything consistent. I pushed one small commit on top to align the review checklist in `references/review.md` with the opt-in wording — hope you don't mind. Merging.

PR #129, opened 2026-07-22
fix: restore agreed token budget wording

Whamp:
## Summary

Follow-up correction to #127.

#127 compressed the reviewed `tokenBudget` schema wording to remain under the existing provider-visible byte ceiling. That was the wrong tradeoff: the agreed language was the requirement, so this PR restores it verbatim and deliberately revises the ceiling instead.

> Optional user-requested soft spend gate, not a planning target. Do not set `tokenBudget` unless the user explicitly supplies a cap or asks you to choose one; never infer or invent one from task size. If omitted, the configured `defaultTokenBudget` applies; without one, the run is unlimited. Reaching the gate blocks later `agent()` calls; concurrent in-flight work can overshoot.

- Pins the full sentence with a strict equality regression test.
- Revises the accepted provider-visible definition ceiling from 4,283 to 4,392 bytes.
- Regenerates the context measurement and guidance baseline.

## Evidence

- `npx tsx --test tests/workflow-tool.test.ts tests/workflow-prompt-budget.test.ts` — 35/35
- `npm run build`
- `npm run docs:check`
- `npm run context:check`
- `npm run guidance:check`
- `npm run delivery-choice -- --model openai-codex/gpt-5.6-sol:high` — 3/3 scenarios passed against the exact wording

### DIFF
--- a/docs/workflow-context-surfaces.json
+++ b/docs/workflow-context-surfaces.json
@@ -13,7 +13,7 @@
     },
     "providerVisibleWorkflowToolDefinition": {
       "serialization": "UTF-8 bytes of JSON.stringify({ name, description, parameters })",
-      "bytes": 4276
+      "bytes": 4392
     },
     "registeredSkillsDiscovery": {
       "serialization": "sum of UTF-8 bytes of normalized Pi skill XML (name + description + location) across every root in package.json's pi.skills",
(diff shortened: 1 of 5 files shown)


QuintinShaw:
Thanks — restoring the exact wording the eval validated is the right call; shipping a paraphrase the harness never saw would quietly invalidate the 3/3 result. I pushed a small commit fixing the comment in the budget test (it credited #127 with the verbatim adoption — that's this PR; #127 was the compression) — hope you don't mind. Merging.

PR #136, opened 2026-07-30
fix: label esc as back in agent detail pager footer (#130)

### LINKED ISSUE #130: Esc key is incorrectly labeled as “summary” in phase details view
After a workflow created, the Workflow Status allows users to inspect individual agents and phases.

When opening a specific phase, such as a Research phase, I can view the execution details for that agent and phase. However, after entering the Summary view, the keyboard shortcut hint at the bottom displays:

↑/↓ line · PgUp/PgDn page · g/G ends · t tail:off · enter summary · esc summary

Both Enter and Esc are labeled as summary.

Esc should represent a Back action and return the user to the previous phase or agent view. This appears to be an incorrect label

**Steps to Reproduce**

Run a Dynamic Workflow.
Open the Workflow Status Panel.
Select an agent or phase, such as a Research phase.
Open the Summary view.
Check the keyboard shortcut hints at the bottom of the panel.

**Actual Behavior**

The shortcut bar shows:
enter summary · esc summary
This makes both keys appear to perform the same action.

**Expected Behavior**

The shortcut bar should show something similar to:
enter summary · esc back
Pressing Esc returned to the previous view.

vansh2408:
## What

Fixes the keyboard shortcut hint in the workflow navigator's agent detail pager footer: `esc` was labeled **summary**, duplicating the `enter` hint. It now reads `enter summary · esc back`.

Fixes #130

## Why

With the pager open, the footer showed `enter summary · esc summary`, making both keys look like they do the same thing and hiding the fact that Esc is the universal "go back one level" key across the navigator. Every other view already labels it `esc back`; this was the one deviation.

## How

- `src/workflow-ui.ts`: changed the `footerHint` pager-open branch label from `"esc summary"` to `"esc back"`. Key handling is untouched - `keyToAction` already maps Esc to `back`, which closes the pager when it is open.
- `tests/workflow-ui-pager.test.ts`: added a test that opens the agent detail pager and asserts the footer shows `enter summary` and `esc back`, and does **not** show `esc summary`, so this regression is now covered.


QuintinShaw:
Thanks — exactly right, and the rendering test asserting the footer content (not just the keymap) is a nice touch. Welcome aboard. Merging.

PR #141, opened 2026-07-31
feat(model): default MiniMax provider to MiniMax-M3

octo-patch:
Reason: The MiniMax provider defaults stop at MiniMax-M2.7; bump both `minimax` and `minimax-cn` defaults to the current MiniMax-M3 model.

## Changes
- `src/model-spec.ts`: updated the `DEFAULT_MODEL_PER_PROVIDER` entries for `minimax` and `minimax-cn` from `MiniMax-M2.7` to `MiniMax-M3`.

These defaults are the model id `buildFallbackModel` selects when a requested model id is not found in the registry for a MiniMax provider, so they are the canonical place the current model is declared.

## Checks
- Verified no tests, docs, generated registries, or guidance snapshots hard-code the previous MiniMax default (a repo-wide search for `MiniMax` only matches the two updated lines).
- No secret material in the diff.


QuintinShaw:
Thanks — verified against the pi-ai model registry: `MiniMax-M3` is live for both the `minimax` and `minimax-cn` providers, and a repo-wide search confirms these two defaults are the only references to the old model. Tests are clean on this change. Merging.

QuintinShaw (approved):

PR #145, opened 2026-08-07
feat(workflow): support named agent threads

rgarcia:
## Problem

I often use workflows where a small-model implementer works in a loop with a large-model reviewer until the reviewer is happy.

Each `agent()` call currently starts with a fresh conversation history. The implementer has to reconstruct its context every turn, which burns tokens and can make review loops slow or fail to converge.

## Change

This adds an optional `thread` to `agent()`. Calls using the same thread name re-enter the same Pi session with its full conversation history and session ID:

```js
const change = await agent('Implement the change.', {
  tier: 'small',
  thread: 'work-session',
})

const review = await agent(`Review the implementation:\n${change}`, {
  tier: 'big',
})

await agent(`Address this review:\n${review}`, {
  tier: 'small',
  thread: 'work-session',
})
```

Threads are scoped to one workflow invocation. Calls on the same thread must be sequential, while unthreaded agents continue to start fresh sessions.

I also considered explicit continuation handles and retaining live `AgentSession` objects. Named threads keep the existing `agent()` return value and only retain the underlying session state needed for re-entry.

## Testing

- `npm test` — 1,194 passed
- [real-provider E2E](https://gist.github.com/rgarcia/30de4db460fd3329f73b858fa5076afa) using `openai-codex/gpt-5.6-sol`
- regression coverage for retries, timeouts, concurrent thread use, worktree isolation, persistence, and workflow resume


QuintinShaw (changes requested):
Thanks @rgarcia — the API shape is clean (keep agent() returning values, re-enter via SessionManager, dispose the AgentSession each turn), and the regression suite around concurrency, worktree rejection, timeout settle-before-retry, journal hash compat, nested sharing, and resume barriers is solid. maintainerCanModify is false on this PR, so I can't push fixes myself.

One real bug before merge:

Turn isolation via messages.length breaks after auto-compaction. In src/agent.ts the threaded path snapshots messagesBeforeTurn = session.messages.length and later does session.messages.slice(messagesBeforeTurn) for empty-output detection and structured lastText. Pi compaction is on by default and can rewrite session.messages inside prompt() after agent_end. After that rewrite the array is much shorter, the slice is empty, finalAssistantText returns "", and the turn is treated as AGENT_EMPTY_OUTPUT — so a turn that actually succeeded gets rolled back, and retries can hit the same wall. That matters for the main use case (multi-turn implementer loops). Length-based indexing is not compaction-safe.

Please key the turn off something stable across compaction — e.g. only inspect messages at/after the last user message when the prompt returns — and cover it with a test that rebuilds/shortens session.messages mid-turn the way compaction does. Apply the same window to the structured-output lastText callback.

Design notes (not merge blockers, but worth confirming):

- Threaded calls as sticky resume barriers means pause/resume starts the thread from scratch and re-runs everything after the first threaded call live. Documented — just want to be sure you're happy with that cost for checkpointy loops.
- Nested workflows sharing bare thread names is powerful for parent→child→parent, and also a silent collision hazard if a child picks the same name independently. Intentional?
- Changing model/tier/agentType across turns on one thread is currently allowed. Intentional?

Happy to re-review once the compaction/turn-window fix is in.

rgarcia:
Thanks for catching the compaction bug — you were right, and it reproduced exactly as described. Fixed in 821dd9e (branch also rebased onto main to clear conflicts).

**The fix.** I went with something slightly stronger than the last-user-message window: the turn's transcript is now collected from `message_end` events on the session (user prompt, assistant messages, tool results, and any schema repair re-prompts all flow through it), and both the empty-output check and the structured `lastText` fallback read that collected window. Index/slice logic on `session.messages` is gone entirely. Two reasons over the user-message anchor:

- In the split-turn case (one turn bigger than `keepRecentTokens`), compaction summarizes away the turn's own user message, so there's no stable user anchor left in the rebuilt array.
- `resolveStructuredOutput`'s repair re-prompts append additional user messages mid-turn; anchoring on the *last* user message would exclude the original response that usually carries the prose payload.

The event window is append-only per turn, so compaction rewriting `session.messages` mid-`prompt()` can't touch it.

**Tests.** Two new integration tests trigger real auto-compaction inside `prompt()` (small `reserveTokens` threshold + a ~54k-token reply, which also exercises the split-turn double-summary path) on the third turn of a named thread — one unstructured, one structured with prose-JSON extraction. Before the fix they fail with exactly `AGENT_EMPTY_OUTPUT` and `SCHEMA_NONCOMPLIANCE`; after it, both turns return their real output. Full suite + release gate green.

**On the design questions:**

- *Resume barriers:* intentional. Replaying a transcript-mutating turn from the journal would hand back a cached result without re-entering the session, desyncing the thread's actual state from what the script believes happened. For checkpointy loops the honest options were "restart the thread live" or "persist full session state into the journal," and the latter didn't seem worth the complexity for v1. Happy with the cost, and it's documented.
- *Nested bare-name sharing:* intentional for parent→child→parent handoffs, but you're right that it's a silent collision hazard for an independently-written child. I've made the guidance explicit in `references/runtime.md`: bare names deliberately re-enter the parent's session; prefix child-internal threads (e.g. `lint:implementer`). If you'd rather have isolation-by-default with an explicit opt-in for sharing, I'm open to that as a follow-up — it's a behavior change I didn't want to fold into this fix.
- *Model/tier changes across turns:* intentional — escalating a stuck implementer thread to a bigger model mid-loop is a real use case, and Pi sessions already support model switching on an existing transcript. Also now documented. Can add a one-time warn log if you'd prefer it loud.

QuintinShaw:
Thanks — collecting the turn from message_end instead of a length snapshot is the right fix, and the two tests really do trip Pi auto-compaction mid-prompt(), including the split-turn path. Happy with the design answers as documented.

QuintinShaw (approved):
Approving — compaction-safe turn window via message_end, with real auto-compaction tests.

QuintinShaw:
Thanks — the compaction turn window was the right call. Merging.

PR #151, opened 2026-08-11
fix: prefer session provider for fuzzy workflow models

tomdale:
## Problem

Workflow `agent()` calls accept bare model names. Before this change, an unqualified fuzzy request could search the full built-in catalog without preferring the active session provider.

For example, in a Pi session using Vercel AI Gateway, this workflow request:

```js
agent("…", { model: "gpt-5.6-luna" });
```

could resolve to `openrouter/openai/gpt-5.6-luna-pro` instead of the configured `vercel-ai-gateway/openai/gpt-5.6-luna`. On a machine without OpenRouter credentials, every affected subagent immediately failed with `No API key found for openrouter`.

## Summary

- prefer the active session provider for unqualified fuzzy workflow model requests
- retain global fuzzy fallback when the preferred provider has no matching model
- cover the Vercel AI Gateway vs OpenRouter GPT-5.6 Luna collision

## Verification

- `npm run test:unit -- --test-name-pattern="prefers the active provider|uses global fuzzy matching"`
- `npm run check:scripts`
- `npm run build`
- `npm run docs:check`
- `npm run context:check`
- `npm run release:verify`

The full unit suite also passed: 1,184 tests.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -775,7 +775,9 @@ export class WorkflowAgent {
     let resolvedModel: Model<any> | undefined;
     let resolvedThinkingLevel: CreateAgentSessionOptions["thinkingLevel"] | undefined;
     if (modelSpec) {
-      const resolved = resolveModelSpecWithThinking(modelSpec, modelRegistry);
+      const resolved = resolveModelSpecWithThinking(modelSpec, modelRegistry, {
+        preferredProvider: this.mainModel?.split("/", 1)[0],
+      });
       if (resolved.warning) console.warn(`[workflow] ${resolved.warning}`);
       if (!resolved.model) {
         if (isExplicitRequest) {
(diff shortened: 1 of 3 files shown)


QuintinShaw:
Thanks @tomdale — this is a real resolution bug, and the fix is scoped the right way.

Unqualified fuzzy matches were free to land on another provider's longer id (OpenRouter's gpt-5.6-luna-pro beating Gateway's gpt-5.6-luna), which blows up when that provider isn't authenticated. Preferring the session provider only for bare names, with global fallback when it has no match, is what I want here. Explicit provider/model pins stay untouched, and the two unit cases cover the important paths.

Approving. Will merge once CI is green.

QuintinShaw (approved):
Approving — bare fuzzy names prefer the session provider, with global fallback. Explicit provider/model pins stay untouched.

QuintinShaw:
Thanks — the provider preference for bare fuzzy names was the right call. Merging.

PR #153, opened 2026-08-14
feat: stream live agent token usage

Whamp:
## What changed

- Stream token estimates while an agent is responding, then replace them with exact provider usage when the response settles.
- Keep provisional usage out of budgets and persisted totals.
- Track concurrent, retried, nested, timed-out, paused, and resumed agents without losing or double-counting usage.
- Preserve the existing once-per-run `onTokenUsage` callback and once-per-agent terminal `onUsage` callback.

## Why

Long-running agents previously showed no token movement until they finished. The live totals now distinguish an active agent from a stalled one without changing final accounting.

## Validation

- `npm run check`
- `npm run build`
- 226 focused runtime, manager, abort, and persistence tests
- Real Pi session streamed 7 updates before settlement, then committed the exact 8,371-token provider total.
- Real Pi timeout returned `AGENT_TIMEOUT` and settled teardown in 522 ms.
- Full suite: 1,071 of 1,072 tests passed. The remaining provider-dependent schema test also fails on unchanged `origin/main` because the model returns an instruction request instead of `{ verdict: "ok" }`.

### DIFF
--- a/src/agent-usage.ts
+++ b/src/agent-usage.ts
@@ -33,10 +33,10 @@ interface AgentCallUsageUpdate {
   committedUsage?: AgentUsage;
 }
 
-/** Settled token count plus exact provider usage when the provider reported it. */
+/** Settled cumulative usage for one logical agent call, including retries. */
 interface AgentUsageCommit {
   tokens: number;
-  terminalUsage?: AgentUsage;
+  tokenUsage?: AgentUsage;
 }
 
 /**
@@ -57,16 +57,14 @@ export function createAgentCallUsageTracker(onUpdate: (update: AgentCallUsageUpd
       const emitProgress = () => {
         onUpdate({ tokenUsage: sumAgentUsage(committedCallUsage, attemptUsage) });
       };
-      const commitUsage = (usage: AgentUsage, reportedTerminalUsage?: AgentUsage): AgentUsageCommit => {
+      const commitUsage = (usage: AgentUsage): AgentUsageCommit => {
         if (!isOpen()) {
           return { tokens: 0 };
         }
         closed = true;
         committedCallUsage = sumAgentUsage(committedCallUsage, usage);
         onUpdate({ tokenUsage: committedCallUsage, committedUsage: usage });
-        return reportedTerminalUsage
-          ? { tokens: usage.total, terminalUsage: reportedTerminalUsage }
-          : { tokens: usage.total };
+        return { tokens: committedCallUsage.total, tokenUsage: committedCallUsage };
       };
 
       return {
@@ -89,7 +87,7 @@ export function createAgentCallUsageTracker(onUpdate: (update: AgentCallUsageUpd
         },
         commitWithFallback(fallbackTotal: number) {
           if (terminalUsage && (terminalUsage.total > 0 || terminalUsage.cost > 0)) {
-            return commitUsage(terminalUsage, terminalUsage);
+            return commitUsage(terminalUsage);
           }
           return commitUsage({ ...createEmptyAgentUsage(), total: Math.max(0, fallbackTotal) });
         },
@@ -98,7 +96,7 @@ export function createAgentCallUsageTracker(onUpdate: (update: AgentCallUsageUpd
             closed = true;
             return { tokens: 0 };
           }
-          return commitUsage(terminalUsage, terminalUsage);
+          return commitUsage(terminalUsage);
         },
       };
     },
--- a/src/agent-usage.ts
+++ b/src/agent-usage.ts
@@ -1,0 +1,118 @@
+/** Token and cost usage for one subagent attempt or logical agent call. */
+export interface AgentUsage {
+  input: number;
+  output: number;
+  cacheRead: number;
+  cacheWrite: number;
+  total: number;
+  cost: number;
+}
+
+/** Create an independent zero-valued agent usage record. */
+export function createEmptyAgentUsage(): AgentUsage {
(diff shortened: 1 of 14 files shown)


QuintinShaw (changes requested):
Thanks @Whamp — the provisional-vs-committed split is the right shape for this. Streaming estimates on the agent, exact settlement into spent/budgets, and keeping onTokenUsage once-per-run is what we need for live panels without lying to accounting. The focused tests around retries, timeout teardown, concurrent/nested identity, and pause/resume settlement are strong.

I can't merge from this head as-is. The branch is based well behind current main (pre #119/#120/#137/#142). A straight conflict resolution that keeps the PR-side workflow/manager code would regress real fixes — notably failed-attempt store.discardDelta, run-fatal sibling abort/drain, and the fail-loud model pin path. Journal identity also diverges (callId here vs runId-namespaced keys on main).

Please rebase onto current main and re-apply the live-usage wiring on top of those semantics, then re-run the focused suite (and a quick real-pi stream smoke if you can). While you're in the stream path, please either document or harden usageFromSessionProgress: today it assumes message_end is emitted before appendMessage so stats+streaming don't double-count — that ordering shouldn't be a silent assumption.

Minor after rebase: prefer terminal/cumulative usage on agentEnd over Math.max with any earlier estimate so a high provisional can't stick.

Happy to re-review once it's rebased.

Whamp:
My bad. Will come back on this.

Whamp:
Addressed the post-rebase usage review in ee56fd8:

- documented and regression-tested the Pi SDK ordering that emits `message_end` before `SessionManager.appendMessage`;
- subtract the named-thread pre-turn usage baseline from live progress, so later turns no longer include earlier turns;
- carry cumulative settled usage across retries into `onAgentEnd`, and let terminal values replace provisional manager snapshots;
- added regressions for threaded live/terminal reconciliation, retry accumulation, and a high provisional estimate settling lower.

Validation: `npm test` passes 1238/1238, all guidance/release checks pass, and a two-turn real Pi smoke on `openai-codex/gpt-5.6-sol:minimal` emitted live usage before settling with each turn’s final live update matching terminal usage.

QuintinShaw (changes requested):
Thanks — the settlement/retry reconciliation is much closer, and the rebase preserved the main-side guards.

One blocking test issue:
- `tests/agent.test.ts:422` and `:453` — the new named-thread live-usage regression never selects its registered faux model. `mainModel` is not the default session model for an untagged run, so in an isolated Node 22 checkout the faux core receives zero calls and the test fails with `AGENT_EMPTY_OUTPUT`. Pass `model: "fauxtest-live-usage/faux-model"` in both calls, or configure a deterministic session default, then rerun the suite.

Everything else I checked is aligned with the requested rebase and settlement semantics.

Whamp:
Addressed in `77df0c4`. Both named-thread test turns now pass `model: "fauxtest-live-usage/faux-model"` explicitly, so the regression no longer depends on the session default.

Verified on Node 22.21.1: the focused regression passes, and the full `npm test` gate passes with 1,238 tests.

QuintinShaw:
Thanks — the explicit model pin closes the test issue. Main moved again while this was in review, and the branch is now conflicting after #162, #163, and #164. Please rebase once more onto current main; I’ll do the final review from that head.

Whamp:
rebased 🤞🏻

QuintinShaw (changes requested):
Thank you for carrying the rebase and settlement work through. The positive path now checks out, including the accounting regressions and a real Pi 0.83.0 / gpt-5.6-luna two-turn named-thread smoke.

One small, bounded abort-path reconciliation gap remains before approval:

- src/agent-usage.ts:94: when an attempt has emitted provisional usage but is paused or aborted without terminal provider usage, commitTerminalUsage() closes the attempt without emitting a rollback to committedCallUsage. The manager therefore retains and persists the abandoned provisional per-agent total even though run-level accounting correctly excludes it. I reproduced this with a runner that reports 100 provisional tokens and rejects on abort without calling onUsage: after pause() settles, the run aggregate is unset as expected, but both the live agent snapshot and persisted agent record still report 100.

The scope is limited: this does not affect finalized run-level budget or cost accounting. It leaves stale per-agent usage in paused or aborted snapshots, persistence, and status-facing displays.

Please reconcile the final display state to the committed usage in the no-terminal branch, or explicitly represent provisional state, and add a manager-level pause or abort regression covering the in-memory agent snapshot, persisted agent record, and status-facing value. The fix looks small, and I am happy to help patch it on the PR branch if useful.

Validation otherwise passed: focused suite 349/349, full suite 1276/1276, guidance check, and the real Luna live and named-thread positive paths.

Whamp:
Addressed the abort-path reconciliation gap in `38bb62d`.

- `commitTerminalUsage()` now emits a rollback to the call's committed usage when an attempt has provisional usage but no terminal provider usage. The rollback carries no `committedUsage` delta, so finalized run accounting remains unchanged.
- Added a manager-level pause regression covering the in-memory agent snapshot, persisted agent record, unset run aggregate, and the navigator/status-facing token value.
- Red/green proof reproduced the stale `100` before the fix and `0` after settlement.

Validation: local `npm test` passes 1,277/1,277 plus the release gate; PR CI is green.

QuintinShaw (approved):
Thanks — 38bb62d closes the abort-path reconciliation gap. The no-terminal branch now rolls the per-agent display state back without committing abandoned usage, and the manager regression covers the live snapshot, persisted record, and status-facing value.

Final re-review passed the focused suite, the full 1,277-test gate, and the guidance and release checks. Approved.

Issue #146, opened 2026-08-07
High priority: maxAgents failure wastes completed-agent tokens and cannot resume with a higher limit


killianMei:
## Severity

**High** — this can irreversibly spend tokens on a run that is guaranteed to fail, then require the same completed work to be paid for again.

## Problem

When a workflow reaches its `maxAgents` ceiling after earlier agents have completed, the next `agent()` call throws `AGENT_LIMIT_EXCEEDED`.

The completed agents have already consumed tokens. The error recommends increasing `maxAgents`, but resume preserves the original limit and the workflow tool's `resumeFromRunId` path does not accept a new `maxAgents`. Starting a new run with a higher limit reruns the completed prefix and spends those tokens again.

## Minimal reproduction

Run this with `maxAgents: 12`:

```js
export const meta = {
  name: 'agent_limit_repro',
  description: 'Reproduce token waste after reaching maxAgents',
}

const results = []
for (let i = 0; i < 12; i++) {
  results.push(await agent(`Return worker result ${i}.`, { label: `worker-${i}` }))
}

return await agent(
  `Synthesize these completed results:\n${results.join('\n')}`,
  { label: 'synthesis' },
)
```

Observed behavior:

1. All 12 worker calls complete and consume tokens.
2. The synthesis call fails with:
   `Agent limit exceeded (12). Use maxAgents option to increase the limit.`
3. Resuming retains `maxAgents: 12`, so the run cannot pass the same point.
4. A new run with `maxAgents: 13` reruns the workers and spends their tokens again.

## Expected behavior

A failed run should be recoverable without repaying for its completed, journaled prefix.

Preferred minimal fix:

- Allow `resumeFromRunId` to explicitly override `maxAgents` for the resumed run, at least when increasing the limit.
- Reuse the existing journaled prefix as normal.
- Update the error message to recommend resume with a higher limit instead of starting a new run.

Possible API shape:

```json
{
  "resumeFromRunId": "<run-id>",
  "script": "<same-or-edited-script>",
  "maxAgents": 20
}
```

## Related work

PR #94 limits wasted calls from an oversized parallel fan-out by cancelling queued agents. That is valuable, but it does not address sequential/dynamic workflows where useful agents have already completed before the cap is reached, nor the inability to raise the cap during resume.

## Environment

- `@quintinshaw/pi-dynamic-workflows` 3.5.1

## Recommended architecture: resumable agent-limit blocking

[...]

PR #160, opened 2026-08-19
feat(runtime): persist checkpoint responses and resume by id

teknologist:
## Problem

Controller-owned checkpoint responses must survive process restart without requiring the workflow's model to echo a large response payload.

## Changes

- Add the durable `checkpoint({ kind, checkpointId, payload })` overload, preserving the legacy prompt overload.
- Persist controller responses through `WorkflowManager.attachCheckpointResponse`, then resume the exact checkpoint ID. No process-global response tokens are required.
- Keep response consumption replay-safe and duplicate resume idempotent; expose checkpoint identity and status without the response payload in control summaries.
- Preserve suspension through parallel/pipeline helpers and script catches, and ensure stop takes precedence while active work drains.
- Document controller attachment, headless behavior, and replay; cover lifecycle and persistence failures.
- Correct the integrated lineage test to assert timestamp ordering rather than same-millisecond event delivery.

## Verification

- `npm test`: 1,456 passed, 0 failed; release gate passed with no warnings.
- Protected guidance reviewed and accepted; documentation, context, and guidance checks passed.
- Real subagent validation before and after integration: pause, persist response, construct a fresh manager, resume, retain the completed journal prefix, and complete the next call. Repeating the same resume did not execute again.
- Independent review completed for the lifecycle fixes and documentation.


QuintinShaw (changes requested):
Thanks — the durable suspend/resume core is carefully done (fail-closed persist, lossless JSON, restart + idempotent consume). The compaction problem is real. I don't think process-global tokens are the right shape for it.

Please rebase onto current main first. resume() conflicts with #146 (raise maxAgents on resume); WorkflowResumeOptions currently drops that field and a naive resolve would regress it.

Then:

- src/workflow.ts:166 / :1642 — do not hardcode kind to "proposal-ready" | "waiting-for-gitlab". This is a general library; keep kind an open identifier.
- Don't add an opaque token store. The response is already persisted on the run at accept time. Let a controller attach the response to the run/checkpoint, and let resume take runId + checkpointId only. The no-model-turn dispatch can stay if you still need it.
- src/workflow-control-tool.ts:293-296 — formatRun/status currently dump the full checkpoint JSON, including the response, back into the model-visible tool result. That undoes the compaction goal even if tokens remain. Surface id/kind/status only.
- If any in-memory handle remains: do not consume it until the checkpoint is actually consumed; a resuming checkpoint that already has a response on disk should be retryable by id after a crash. Also needs a TTL — stale/conflict paths never release the map entry today.
- src/checkpoint-response-token.ts:97-133 — the BroadcastChannel "bridge" has no sender, no auth, and cannot share the token map across workers. Please drop it.

Happy to re-review after the rebase.

QuintinShaw:
Rechecked the current head (`e6a2bf2`) during the latest maintenance pass. The earlier changes request still applies: the process-global token map and BroadcastChannel dispatch remain, the response is still included in model-visible status output, and checkpoint kinds remain application-specific.

I am leaving this PR open with changes requested, rather than merging the current API. Please rework it around a durable response attached to the run/checkpoint, identifier-only resume/status, general checkpoint kinds, and restart/idempotency coverage on current main. The earlier review contains the detailed acceptance points; the underlying compaction use case remains valid.

teknologist:
@QuintinShaw, I rewrote this PR onto current `main` (`3a6c259`) and pushed signed commit `7df356d`.

This revision addresses the review points:

- It removes the process-global token map and `BroadcastChannel`. The feature is rebuilt on current `main`, while preserving the existing `maxAgents` resume behavior.
- `await manager.attachCheckpointResponse(runId, checkpointId, response)` now persists the response under the run lease, waits for the local execution to settle, and synchronizes active state.
- `workflow_control` resumes a checkpoint with identifiers only. Checkpoint status exposes only `id`, `kind`, and `status`; it does not expose the response.
- Checkpoint `kind` remains an open identifier instead of an application-specific union.
- Restart, early attachment, concurrent-conflict, and post-consumption crash replay cases now have focused coverage. A response that is already on disk remains retryable by identifier after a crash.
- A separate process that tries to attach while the run is busy must retry after the run lease becomes available.

Verification:

- Independent code-auditor review approved the revision after fixes for the attachment race and post-consumption crash replay.
- 161 focused checkpoint, manager, and control tests pass.
- Build, `docs:check`, `context:check`, `guidance:check`, `release:verify`, Biome, and diff checks pass.
- The full suite is not green: 1,440 tests ran, 1,436 passed, and 4 failed in unchanged `tests/agent.test.ts` (session-directory fallback, two compaction cases, and model fallback). All four failures reproduce on pristine base `3a6c259` in the same environment (`npx tsx --test tests/agent.test.ts`: 100 tests, 96 passed, 4 failed). A second environment reproduced two of them.
- I did not run a real-provider end-to-end test.

The PR head is now `7df356dc9969a21d7fcf3585e1a1435a142d181f`.

QuintinShaw:
Thanks for reworking this around persisted responses. Merged after fixing two additional lifecycle cases: helper-level recovery could swallow checkpoint suspension, and stop during active-work drain could be overwritten as paused. Both now have regressions, and the durable controller handoff is documented. The final integrated suite passed all 1,456 tests, the latest CI is green, and real subagent tests verified response attachment across restart, unchanged-prefix replay, continuation, and idempotent duplicate resume.

PR #167, opened 2026-08-25
feat: show a running agent's real model in /workflows

### LINKED ISSUE #166: A running subagent displays the main session model, not the model it actually runs on
### What happens

While a workflow subagent is **running**, `/workflows` (and the detailed task panel) shows the **main session model** for it, not the model the subagent actually runs on. The value only becomes correct once the agent finishes.

Reproduced end-to-end against real subagent sessions on v3.8.0 (`~/.pi/agent/settings.json` default `openai-codex/gpt-5.6-sol`, `~/.pi/workflows/model-tiers.json` = `{small: …:medium, medium: …:high, big: …:max}`):

```
  start  untagged  model=openai-codex/gpt-5.6-sol        <- displayed for the whole run
  end    untagged  model=openai-codex/gpt-5.6-sol:high   <- only now correct

  start  tier-big  model=openai-codex/gpt-5.6-sol        <- displayed for the whole run
  end    tier-big  model=openai-codex/gpt-5.6-sol:max    <- only now correct
```

### Why

`agent()` seeds the display value before it can possibly know the answer:

```ts
// src/workflow.ts:700
let displayModel = modelSpec ?? options.mainModel;
// src/workflow.ts:741 — emitted before agentRunner.run() is ever called
options.onAgentStart?.({ …, model: displayModel });
```

The real model is picked inside `WorkflowAgent.run()` and reported at `src/agent.ts:914`, but the workflow-side handler only mutates the closure local:

```ts
// src/workflow.ts:847
onModelResolved: (id: string) => { displayModel = id; },
```

`agent.model` is assigned in exactly two places in the package — `WorkflowManager`'s `onAgentStart` handler (`src/workflow-manager.ts:858`) and its `onAgentEnd` handler (`:906`) — and `emitLive` has no model-carrying event in between, so the snapshot field is physically unwritable while the agent runs. Every render surface (`src/task-panel.ts:941`, `src/workflow-ui.ts:1055/1472/1497`) reads that one field, so they are all wrong together and all flip at completion.

### Why it is not an edge case

Two very common shapes hit it:

1. **A `tier`-tagged agent.** `src/workflow.ts:696` deliberately passes `modelSpec: undefined` when `opts.tier` is set, so `displayModel` *always* degrades to `mainModel`. `/code-review`'s own agents are tier-tagged.
2. **Any untagged agent, once `model-tiers.json` exists.** `resolveAgentModelSpec` (`src/agent.ts:190-209`) implicitly routes untagged agents through the `"medium"` tier. So configuring tiers at all makes every agent in every workflow display the wrong model while running.

### A worse variant: resume never converges

The journal-replay branch (`src/workflow.ts:767/774`) emits `onAgentStart` **and** `onAgentEnd` back-to-back with the never-resolved value, and `JournalEntry` records no model. So after a pause/resume, replayed rows show the main model **permanently** — combined with `agents: []` on resume (`src/workflow-manager.ts:1380`), that regresses rows which were correct before the pause.

Related but distinct: an agent aborted mid-run never reaches `onAgentEnd` at all, so its stale model is persisted forever.

### Notes

- Not a display-layer bug — the value never reaches the snapshot, so no repaint can heal it.
- `README.md:83` advertises live tracking of "phases, agents, models …", and `src/display.ts:27` documents the field as "the model this agent **ran on**", so the current behaviour contradicts both.
- Closest prior art is #148 (stale `mainModel`, fixed by #155) — which is why the value shown is the *current* main model rather than a stale one.

PR follows.

Yuanpeng-Li:
Fixes #166.

While a subagent is **running**, `/workflows` shows the **main session model** instead of the model it actually runs on; the display only becomes correct once the agent finishes. See #166 for the full analysis — short version: `onAgentStart` is necessarily emitted before `agentRunner.run()`, so it can only carry a pre-resolution guess (`modelSpec ?? options.mainModel`), and `onModelResolved` (src/workflow.ts:847) only mutated a closure local. `agent.model` had exactly two writers — the manager's `onAgentStart` and `onAgentEnd` handlers — so the snapshot field was physically unwritable mid-run.

This hits **tier-tagged agents always** (`agent()` deliberately passes `modelSpec: undefined` when `opts.tier` is set) and **every untagged agent** once `model-tiers.json` exists (`resolveAgentModelSpec` implicitly routes untagged agents through the `"medium"` tier).

## What this does

1. **`WorkflowRunOptions.onAgentModel`** (`src/workflow.ts`) — new *optional* callback, fired from the existing `onModelResolved` hook with the same per-call `id` as `onAgentStart`/`onAgentEnd`/`onAgentUsage`/`onAgentHistory`.
2. **`WorkflowManager`** — handler writes `agent.model` into the live snapshot and re-emits `"agentModel"`. Shaped exactly like the existing `agentHistory` emit (`{ runId, agentId, ...event }`). No `persistRun()` — the throttled progress persist and every terminal persist already pick the field up.
3. **Repaint** — `"agentModel"` added to `RUN_EVENTS` (`src/task-panel.ts`) and the navigator's `events` list (`src/workflow-ui.ts`). `workflow-commands.ts`'s `progressEvents` is deliberately untouched: that surface renders no model.
4. **Resume** — `JournalEntry` / `PersistedJournalEntry` gain an optional `model`, recorded at `onAgentJournal` and used by the replay branch, so replayed rows stop showing the main model forever. Entries without it degrade to exactly the old behaviour.

Side benefit: an agent aborted mid-run never reaches `onAgentEnd`, so its stale model used to be persisted forever; it is now already correct when the abort lands.

## Deliberately not done

- **`model` is not added to `hashAgentCall`.** The cache key stays spec-level, so **no cached agent is invalidated** and no paid work is replayed. The journaled model is a record of what ran, not an identity input.
- **No eager `resolveAgentModelSpec` in `workflow.ts`.** It would duplicate the precedence ladder from `src/agent.ts:190-209`, could not canonicalize a fuzzy spec (`"opus"`, `"gpt-5.5:xhigh"`) without the registry that is only built asynchronously at `src/agent.ts:858`, and would actively *lie* in the untagged-degrade path at `src/agent.ts:900-910`, where the requested tier model is unavailable and the session default is used instead.
- **`onAgentStart` is not moved after resolution.** `onModelResolved` lives inside `if (modelSpec)`, so on a fresh install with no tiers configured it never fires — the agent row would never be created at all. It would also corrupt per-agent `startedAt` and let `onAgentUsage`/`onAgentHistory` arrive before the row exists, where both are silently dropped.
- **No emit on the `onModelFallback` path.** It is reachable only for an untagged agent, and in exactly that case `displayModel` is already `options.mainModel` — the session default it degrades to — so the event would carry no new information. Happy to add it with a `degraded: true` flag if you'd prefer the UI to mark it.

## Back-compat

Optional callback, so no embedder implementing `WorkflowRunOptions` breaks. `WorkflowAgentRunner` is untouched — a custom runner that never calls `onModelResolved` simply keeps the old display. Old journals on disk degrade via `cached.model ?? displayModel` (covered by a test); an old build reading a new journal ignores the extra key.

One deliberate semantic: if the user edits `model-tiers.json` between pause and resume, a replayed row shows the model that **actually produced the cached result**, not a fresh guess. That is strictly more accurate than the old behaviour, which showed whatever `mainModel` happened to be at resume time.

## Verification

`npm test` — Biome, tsc, build, `docs:check`, `context:check`, `release:verify` all green; **1283/1283** unit tests pass.

No generated surface needed regenerating: `onAgentModel` adds no workflow-script capability, no `workflow` tool parameter and no `~/.pi/workflows/settings.json` key, so the capability contract, authoring reference, skills and the README config sections are unaffected (`docs:check` / `context:check` / `guidance:check` all confirm "unchanged"). `README.md:83`'s existing claim that the panel tracks models becomes true rather than needing an edit.

**Six new tests**, each mutation-verified (reverting the corresponding line makes exactly that test fail):

- `onAgentModel` fires with the resolved id and lands **before** `onAgentEnd` (asserts the `start → model → end` order);
- same for a `tier`-tagged agent, and that a tier is still not pre-resolved into an explicit `model` (guards the `src/workflow.ts` half of the contract);
- a replayed cache hit reports the journaled model, and the runner is never called;
- a legacy journal entry with no `model` degrades to the old behaviour;
- the live snapshot is corrected while the agent is still inside `run()`, the `"agentModel"` event is broadcast, and the corrected value is persisted;
- `installTaskPanel` repaints on `"agentModel"` and unsubscribes on dispose (pins `RUN_EVENTS` membership, which nothing previously covered).

**End-to-end against real subagent sessions** (per CONTRIBUTING — real `createAgentSession`, real model, no fake agent; throwaway harness deleted before commit). Session model `openai-codex/gpt-5.6-sol`, tiers `{small: …:medium, medium: …:high, big: …:max}`:

```
--- LIVE RUN (real subagent sessions) ---
  start  untagged  model=openai-codex/gpt-5.6-sol
  MODEL  untagged  model=openai-codex/gpt-5.6-sol:high   <- pushed WHILE RUNNING
  end    untagged  model=openai-codex/gpt-5.6-sol:high  result="pong"
  start  tier-big  model=openai-codex/gpt-5.6-sol
  MODEL  tier-big  model=openai-codex/gpt-5.6-sol:max    <- pushed WHILE RUNNING
  end    tier-big  model=openai-codex/gpt-5.6-sol:max   result="pong."

journaled models: ["openai-codex/gpt-5.6-sol:high","openai-codex/gpt-5.6-sol:max"]

--- RESUME / REPLAY (cache hits, no model resolution happens) ---
  start  untagged  model=openai-codex/gpt-5.6-sol:high
  end    untagged  model=openai-codex/gpt-5.6-sol:high
  start  tier-big  model=openai-codex/gpt-5.6-sol:max
  end    tier-big  model=openai-codex/gpt-5.6-sol:max
```

Also verified by probe harness (deleted): retries re-emit with the same `deltaKey` (idempotent overwrite, journal and `onAgentEnd` agree on the last value); named threads get one emit per turn on that turn's own row and correctly never journal a model; nested `workflow()` inherits the callback through the `...options` spread and its ids stay namespaced by `${runId}-nested${n}`; `checkpoint()` journal entries legitimately carry no model and cannot collide with an agent entry because the hashes differ.

## Typed `feat:`, not `fix:`

Per CONTRIBUTING — it adds a new exported option (`WorkflowRunOptions.onAgentModel`) — even though its purpose is a bug fix.

## Unrelated pre-existing flake (not from this PR)

`tests/workflow-manager-abort.test.ts` → "external signal registration and cleanup failures converge without replacing execution outcomes" fails intermittently. It fails at the same rate on unmodified `origin/main` (3/6 runs there, 4/6 on this branch). Cause: the test creates five managers sharing one `cwd` and then asserts on `listRuns()[0]`, which `RunPersistence.list()` sorts by `updatedAt` **descending** (`src/run-persistence.ts:364`) — when two runs land in the same millisecond the ordering flips and `[0]` is a different manager's completed run. Left alone here to keep this PR to one concern; happy to send a separate PR keying those assertions by `runId`.

### DIFF
--- a/src/run-persistence.ts
+++ b/src/run-persistence.ts
@@ -51,6 +51,8 @@ export interface PersistedJournalEntry {
   hash: string;
   result: unknown;
   storeDelta?: Record<string, unknown>;
+  /** The model the call ran on; absent on journals written before this field existed. */
+  model?: string;
 }
 
 export interface PersistedRunState {
--- a/src/task-panel.ts
+++ b/src/task-panel.ts
@@ -36,6 +36,7 @@ import { shortModel } from "./workflow-ui.js";
 // it redraws identical content.
 const RUN_EVENTS = [
   "agentStart",
+  "agentModel",
   "agentEnd",
   "phase",
   "log",
--- a/src/workflow-manager.ts
+++ b/src/workflow-manager.ts
@@ -869,6 +869,20 @@ export class WorkflowManager extends EventEmitter {
           this.emitLive(managed, "agentStart", { runId: managed.runId, ...event });
           progress();
         },
+        onAgentModel: (event) => {
+          // The ONLY mid-run correction of a running agent's displayed model.
+          // agentsById is keyed by the per-CALL id (never the label), so a
+          // concurrent same-label sibling can't be misattributed.
+          const agent = managed.agentsById.get(event.id);
+          if (!agent) {
+            return;
+          }
+          agent.model = event.model;
+          // No persistRun() here: this fires once per attempt, and the throttled
+          // progress persist plus every terminal persist already pick the field up.
+          this.emitLive(managed, "agentModel", { runId: managed.runId, agentId: agent.id, ...event });
+          progress();
+        },
         onAgentUsage: (event) => {
           const agent = managed.agentsById.get(event.id);
           if (!agent) {
--- a/src/workflow-ui.ts
+++ b/src/workflow-ui.ts
@@ -1946,6 +1946,7 @@ export function openWorkflowNavigator(
       const renderCache = new NavigatorTextRenderCache();
       const events = [
         "agentStart",
+        "agentModel",
         "agentEnd",
         "phase",
         "log",
--- a/src/workflow.ts
+++ b/src/workflow.ts
@@ -84,6 +84,13 @@ export interface JournalEntry {
    * which agent finished first. Absent on older journal entries.
    */
   storeDelta?: Record<string, unknown>;
+  /**
+   * The model this call actually ran on, captured post-resolution so a replayed
+   * cache hit displays what really ran instead of the pre-resolution guess.
+   * Absent on journal entries persisted before this field existed (and on
+   * checkpoints, which run no model) — those degrade to the old behavior.
+   */
+  model?: string;
 }
 
 /**
@@ -266,6 +273,19 @@ export interface WorkflowRunOptions extends WorkflowAgentOptions {
     committedUsage?: AgentUsage;
   }) => void;
   onAgentHistory?: (event: { id: string; label: string; phase?: string; history: AgentHistoryEntry[] }) => void;
+  /**
+   * The agent's REAL model, pushed the moment WorkflowAgent resolves it — mid-run,
+   * long before onAgentEnd. onAgentStart can only carry the pre-resolution guess
+   * (this call's explicit/phase spec, else the session's main model), which is wrong
+   * for every tier-routed agent: an explicit `tier` deliberately defers the choice to
+   * the agent layer, and an untagged agent is implicitly routed through the "medium"
+   * tier whenever model-tiers.json exists (see resolveAgentModelSpec). Without this
+   * channel those agents display the main session model for their whole lifetime and
+   * only flip to the truth once they finish. Fires once per ATTEMPT (and per turn for
+   * a named thread), so treat it as idempotent, not once-per-agent. `id` is the same
+   * per-CALL id as onAgentStart/onAgentEnd/onAgentHistory/onAgentUsage.
+   */
+  onAgentModel?: (event: { id: string; label: string; phase?: string; model: string }) => void;
   onTokenUsage?: (usage: {
     input: number;
     output: number;
@@ -694,9 +714,12 @@ export async function runWorkflow<T = unknown>(
     const explicitModel = agentOptions.model ?? agentDef?.model;
     const modelSpec =
       explicitModel ?? (agentOptions.tier ? undefined : resolveModelForPhase(assignedPhase, routingConfig));
-    // For display in /workflows: the model this agent runs on — its explicit/phase
-    // spec, else the session's main model. The real resolved id overrides this via
-    // onModelResolved once the subagent session is created.
+    // For display in /workflows: a PRE-RESOLUTION guess — this agent's explicit/phase
+    // spec, else the session's main model. It is only a guess: a `tier` deliberately
+    // leaves modelSpec undefined so the agent layer picks, and an untagged agent is
+    // implicitly routed through the "medium" tier when model-tiers.json exists. The
+    // real resolved id replaces it via onModelResolved below, which also pushes the
+    // correction out on onAgentModel so a RUNNING agent's row stops showing the guess.
     let displayModel = modelSpec ?? options.mainModel;
 
     // Deterministic resume key: assigned at lexical call time, before the limiter,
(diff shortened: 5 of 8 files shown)


QuintinShaw (approved):
Thanks — the model-resolution boundary and live update path look correct. The resolved model now reaches the running snapshot through opaque call identity, persisted results retain the model that produced them, and legacy journals remain compatible without changing cache identity. Focused and full local gates passed, along with a real tier-routed Pi smoke. Approved.

PR #168, opened 2026-08-28
feat: emit background workflow lifecycle events

Ancientkingg:
Other Pi extensions currently learn about a background workflow only when its conversation result is delivered. That is too late for integrations such as terminal activity indicators.

This adds `WORKFLOW_LIFECYCLE_EVENT` on `pi.events`. Background runs emit `started`, `resumed`, `paused`, `completed`, `failed`, and `stopped` with the run ID, workflow name, and session ID. Retained managers refresh the emitter after `/reload`, so events go through the current extension runtime.

Adds focused manager and reload regression tests. `npm test` passes.


src/task-panel.ts
```diff
@@ -720,6 +731,30 @@ export function installResultDelivery(
...
   patchBindCoreObserve();
 
+  m.__lifecycleEventEmitter = (data) => pi.events?.emit(WORKFLOW_LIFECYCLE_EVENT, data);
+  if (!m.__lifecycleEventInstalled) {
+    m.__lifecycleEventInstalled = true;
+    const emitLifecycle =
+      (status: WorkflowLifecycleEvent["status"]) =>
+      ({ runId }: { runId: string }) => {
+        const run = manager.getRun(runId);
+        if (!run?.background) return;
```
QuintinShaw: **[bug]** `emitLifecycle` bails out unless `manager.getRun(runId)` returns a live in-memory background run. That is correct for `started`/`resumed`/`paused`/`complete`/`error` (those fire while the run is still in `this.runs`), but `WorkflowManager.stop()` has an explicit cold-start path: when the run is only on disk (`running`/`paused` after process restart, version-mismatch pause, or `recoverStaleRuns()`), it still `emit("stopped", { runId })` after rewriting the persisted row to `aborted`. `getRun(runId)` is `undefined` there, so `!run?.background` treats a real background stop as "not a background run" and never publishes `WORKFLOW_LIFECYCLE_EVENT`. This is a user-facing action other extensions are documented to observe; the existing `cold-start stop` manager tests already construct this case.

**Suggestion:** When `getRun()` misses, fall back to `manager.getPersistence()?.load(runId)` (or `listAllRuns()`) for `workflowName`/`sessionId`. A disk-only stoppable row is not an in-process foreground `runSync` call, so it should still emit `{ status: "stopped", runId, name, sessionId? }` without reading `script`/`args`. Add a test that `installResultDelivery` plus a real manager `stop()` of a persisted-only paused/running run delivers exactly one `stopped` event, and that `/reload` still routes it through the fresh `pi.events`.

README.md
```diff
@@ -252,7 +252,7 @@ Extension state lives outside the repository under `~/.pi/workflows`:
 
 Subagents are in-memory by default. Set `persistAgentSessions: true` to retain full transcripts in Pi's standard session directory. This creates one file per unthreaded call or named thread and may store sensitive material that an agent read, so enable it deliberately.
 
-Completed background runs persist their full result in the project run JSON. The conversation delivery includes a pointer to that file when the visible summary is shortened.
+Completed background runs persist their full result in the project run JSON. The conversation delivery includes a pointer to that file when the visible summary is shortened. Other Pi extensions can subscribe to the exported `WORKFLOW_LIFECYCLE_EVENT` through `pi.events`. Background workflows emit `{ status, runId, name, sessionId }`, where `status` is `started`, `resumed`, `paused`, `completed`, `failed`, or `stopped`.
```
QuintinShaw: **[nit]** The README presents the payload as `{ status, runId, name, sessionId }`, but `WorkflowLifecycleEvent.sessionId` is optional and the emitter omits the field when `resolveDeliverySessionId` returns nothing (unbound manager / pre-`session_start`).

**Suggestion:** Document `sessionId` as optional, e.g. "emits `{ status, runId, name }` and includes `sessionId` when the originating Pi session is known."

QuintinShaw (changes requested):
## Summary

The PR adds a small, well-scoped `pi.events` bridge for background workflow lifecycle (`started`/`resumed`/`paused`/`completed`/`failed`/`stopped`), keeps the payload to identifiers only (no script/args/result/error text), and correctly refreshes the emitter on `/reload` so retained managers do not double-subscribe or keep talking to a stale `ExtensionAPI`. Manager-side `started` is emitted only after the initial persist, existing complete/error/pause/stop paths already fire before in-memory eviction, and tests cover the reload emitter swap plus the new `started` event. The one correctness gap is that `stop()`'s documented disk-only fallback still emits on the manager, but the new bridge drops it because it requires a live `getRun()` entry.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 1

src/task-panel.ts
Ancientkingg: Fixed in f73e67c. The lifecycle bridge now falls back to persisted run metadata when no live run exists, so cold-start stops emit once through the current `pi.events` runtime. I added a regression test using a real manager with a persisted-only paused run and a reload emitter swap.

README.md
Ancientkingg: Updated in f73e67c. The README now documents `{ status, runId, name }` as the base payload and notes that `sessionId` is included when the originating session is known.

QuintinShaw (approved):
Re-reviewed `f73e67c` against the previous CHANGES_REQUESTED items.

Both are fixed in source:

1. `emitLifecycle` falls back to persisted run metadata when `getRun` is empty, so a cold-start `stop()` of a disk-only `running`/`paused` run emits `stopped` once, with `name`, through the current `pi.events` runtime. Live foreground runs still do not emit. The new regression test drives a real manager with a persisted-only paused run and a reload emitter swap.
2. README now documents `{ status, runId, name }` and notes that `sessionId` is included only when the originating session is known.

`test` is green on this head.

This branch is still based on v3.9.0. `workflow-manager.ts` also changed on `main` in #172, so update the branch before merge (strict required checks). The lifecycle payload stays identifiers-only.

PR #169, opened 2026-08-28
fix(delivery): auto-deliver background workflow results in omp

donrami:
## Problem

In omp (the bundled fork of pi), background workflow runs never delivered their results back into the host conversation. Completed runs sat in `pendingDelivery` on disk forever.

## Root cause

The delivery engine needs a **session-stable, thenable send** (`AgentSession.sendCustomMessage`) captured from the host. omp's quiet sessions never call `sendCustomMessage` themselves, so the capture patch never fired and every bind stayed fail-closed.

## Fix

- **Bind-time capture probe** (`src/task-panel.ts`): `bindSessionDelivery` sends one invisible, no-turn `workflow-delivery-probe` message via `pi.sendMessage`; the patched `sendCustomMessage` captures the live host session from its receiver. The wrapper treats the probe as **capture-only** — it never forwards to the real send, so nothing is appended, persisted, or turned.
- **Retry semantics**: the probed-set is marked only after a send is actually captured, so a failed probe retries on the next bind; `dropSessionDelivery` clears the mark.
- **Persistence gate**: omp persists the host session lazily (after bind), so the on-disk gate does not reject probe-bearing sends. Child sessions (`workflow:`-named) still can't capture.
- **Registry fallback** (`src/agent.ts`): the fallback registry resolves through omp's `discoverAuthStorage`, legacy `ModelRegistry.create`, or `ModelRuntime.create` — a real `ModelRegistry` on every host, rejection never cached. Subagent handoff spreads `modelRuntime` when available and falls back to `modelRegistry` for omp's fork, never an `undefined` runtime.

## Verification

- 1312 tests (biome + tsc clean); 2 pre-existing upstream failures on `2f85c9e` (`workflow-authoring-skill`, `workflow-release-gate`) reproduce on pristine upstream — unrelated.
- Live E2E on **omp** (bundled fork): background run delivered a `workflow-result` turn (`triggerTurn: true`, `deliverAs: followUp`), `pendingDelivery` cleared, **no probe entry persisted** in the transcript.
- Live E2E on **stock pi 0.84.3** (extension mounted via `-e`, separate auth): same delivery, zero probe pollution.
- Retried-probe and drop/rebind semantics covered by dedicated regression tests.

### DIFF
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -10,6 +10,7 @@ import {
   DefaultResourceLoader,
   getAgentDir,
   ModelRegistry,
+  type ModelRuntime,
   SessionManager,
   SettingsManager,
   type ToolDefinition,
@@ -251,32 +252,51 @@ export interface WorkflowAgentOptions {
 // runtime split): build it directly from the discovered AuthStorage. The
 // disk-backed fallback is built lazily; sync callers see [] until it resolves
 // and real specs on later reads.
-let fallbackRuntimePromise: Promise<ModelRegistry | undefined> | undefined;
+let fallbackRuntimePromise: Promise<ModelRegistry> | undefined;
 let fallbackRegistry: ModelRegistry | undefined;
 
 function ensureFallbackRegistry(): Promise<ModelRegistry> {
   if (!fallbackRuntimePromise) {
     const dir = getAgentDir();
     // Same auth.json/models.json createAgentSession uses by default, so a model
-    // resolved here carries valid credentials.
+    // resolved here carries valid credentials. Three host shapes, tried in
+    // order; only a real registry ever resolves:
+    // 1. omp's bundled pi exports discoverAuthStorage (auth-storage-backed
+    //    registry, no runtime split).
+    // 2. Legacy pi (< 0.80.8) had a static ModelRegistry.create({ dir }).
+    // 3. pi >= 0.80.8 splits registry/runtime: ModelRuntime.create() then
+    //    new ModelRegistry(runtime).
     fallbackRuntimePromise = (async () => {
-      // discoverAuthStorage is omp's bundled pi-coding-agent export; on other
-      // hosts (pi CLI, tests) fall back to the legacy static create, and if
-      // neither exists the fallback registry stays empty (listAvailableModels
-      // reports [] and tier ranking falls through to the main model).
-      const mod = await import("@earendil-works/pi-coding-agent").catch(() => undefined);
-      const discover = (mod as { discoverAuthStorage?: (dir: string) => unknown } | undefined)?.discoverAuthStorage;
-      if (typeof discover === "function") {
-        const authStorage = (await discover(dir)) as ConstructorParameters<typeof ModelRegistry>[0];
+      // Dynamic import on purpose: discoverAuthStorage exists only in omp's
+      // bundled pi; a static named import would fail to resolve on stock pi,
+      // and the host shape must be feature-detected at runtime.
+      const ompExports = (await import("@earendil-works/pi-coding-agent").catch(() => undefined)) as
+        | {
+            discoverAuthStorage?: (dir: string) => Promise<unknown>;
+            ModelRuntime?: { create?: (options?: unknown) => Promise<ModelRuntime> };
+          }
+        | undefined;
+      if (typeof ompExports?.discoverAuthStorage === "function") {
+        const authStorage = (await ompExports.discoverAuthStorage(dir)) as ConstructorParameters<
+          typeof ModelRegistry
+        >[0];
         return new ModelRegistry(authStorage);
       }
-      const create = (
+      const legacyCreate = (
         ModelRegistry as unknown as {
           create?: (opts: { dir: string }) => Promise<ModelRegistry>;
         }
       ).create;
-      if (typeof create === "function") return create({ dir });
-      return undefined;
+      if (typeof legacyCreate === "function") return legacyCreate({ dir });
+      const runtimeCreate = ompExports?.ModelRuntime?.create;
+      if (typeof runtimeCreate === "function") {
+        // No options: defaults to getAgentDir()/auth.json and models.json —
+        // the same disk layout the omp and legacy paths read.
+        return new ModelRegistry(await runtimeCreate());
+      }
+      throw new Error(
+        "[workflow] no ModelRegistry construction path in @earendil-works/pi-coding-agent (expected discoverAuthStorage, ModelRegistry.create, or ModelRuntime.create)",
+      );
     })();
     // Don't cache a rejection: a transient failure (e.g. auth.json lock) would
     // otherwise wedge the fallback for the rest of the process.
@@ -286,17 +306,17 @@ function ensureFallbackRegistry(): Promise<ModelRegistry> {
   }
   return fallbackRuntimePromise.then((registry) => {
     fallbackRegistry ??= registry;
-    return fallbackRegistry;
-  }) as Promise<ModelRegistry>;
+    return registry;
+  });
 }
 
 /**
  * The ModelRuntime behind a registry facade (pi >= 0.80.8 shape: ModelRegistry
  * wraps a ModelRuntime but exposes no getter). Subagent sessions need it to
- * share the host session's exact catalog and auth — createAgentSession takes
- * modelRuntime, not a registry, on that line. omp's fork is auth-storage-backed
- * (registry has no `runtime` field and createAgentSession takes modelRegistry
- * instead), so this returns undefined there and the handoff no-ops.
+ * share the host session's exact catalog and auth. omp's fork is
+ * auth-storage-backed (registry has no `runtime` field and createAgentSession
+ * takes modelRegistry instead), so this returns undefined there and callers
+ * pass the registry itself.
  */
 export function runtimeOf(registry: ModelRegistry): unknown {
   return (registry as unknown as { runtime?: unknown }).runtime;
@@ -937,6 +957,13 @@ export class WorkflowAgent {
     // temporary worktree paths.
     const sessionManager = this.createSessionManager(options.thread);
     const threadLeaf = options.thread ? sessionManager.getLeafId() : null;
+    // Host split: pi >= 0.80.8 createAgentSession takes modelRuntime, not a
+    // registry — hand over the registry's backing runtime so subagents share
+    // the host catalog and auth. omp's fork is auth-storage-backed (registry
+    // has no `.runtime`; it takes modelRegistry instead). Never spread
+    // `modelRuntime: undefined` — it would shadow createAgentSession's own
+    // default runtime.
+    const modelRuntime = runtimeOf(modelRegistry) as ModelRuntime | undefined;
     let session: Awaited<ReturnType<typeof createAgentSession>>["session"];
     try {
       ({ session } = await createAgentSession({
@@ -954,12 +981,10 @@ export class WorkflowAgent {
         // wins and skips the shared build entirely; the ...this.sessionOptions
         // spread below re-applies the same injected value harmlessly.
         resourceLoader: this.sessionOptions.resourceLoader ?? (await this.getSharedResourceLoader(agentDir)),
-        // pi >= 0.80.8: createAgentSession takes modelRuntime, not a registry
-        // (the registry option is ignored there). omp's fork has no `.runtime`,
-        // so this spread is a no-op on omp.
-        ...(modelRegistry
-          ? { modelRuntime: runtimeOf(modelRegistry) as CreateAgentSessionOptions["modelRuntime"] }
-          : {}),
+        // Host split (see modelRuntime above): stock pi takes modelRuntime;
+        // omp's fork takes modelRegistry. Spread-cast keeps the runtime value
+        // while satisfying the upstream CreateAgentSessionOptions type.
+        ...(modelRuntime ? { modelRuntime } : { modelRegistry }),
         ...this.sessionOptions,
         // The computed AgentRegistry id must win over any injected
         // sessionOptions value: a stable embedder-supplied agentId would
--- a/src/agent.ts
+++ b/src/agent.ts
@@ -10,7 +10,6 @@ import {
   DefaultResourceLoader,
   getAgentDir,
   ModelRegistry,
-  ModelRuntime,
   SessionManager,
   SettingsManager,
   type ToolDefinition,
@@ -220,7 +219,7 @@ export interface WorkflowAgentOptions {
    * so a workflow subagent can't fan out through them either (#107).
    */
   excludeTools?: string[];
-  /** Override any createAgentSession option (model, modelRuntime, resourceLoader, etc.). */
+  /** Override any createAgentSession option (model, modelRegistry, resourceLoader, etc.). */
   session?: Partial<CreateAgentSessionOptions>;
   /** Extra system guidance prepended to every subagent task. */
   instructions?: string;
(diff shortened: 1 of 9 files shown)


src/task-panel.ts
```diff
@@ -209,6 +209,35 @@ const boundSessionSends = new Map<string, DeliverySend>();
...
+ * live session's sendCustomMessage synchronously, the prototype patch captures
+ * the receiver, and boundSessionSends is populated by the time this returns.
+ * display:false + triggerTurn:false appends a durable invisible entry on every
+ * supported host and starts no turn.
+ */
+function probeHostSessionSend(pi: ExtensionAPI, sessionId: string): void {
+  if (probedSessionIds.has(sessionId)) return;
+  probedSessionIds.add(sessionId);
+  try {
+    pi.sendMessage({ customType: DELIVERY_PROBE_CUSTOM_TYPE, content: "", display: false }, { triggerTurn: false });
```
QuintinShaw, line 243: **[bug]** `probeHostSessionSend` calls `pi.sendMessage({ customType: "workflow-delivery-probe", content: "", display: false }, { triggerTurn: false })`, and `patchedSendCustomMessage` always forwards to the original. On pi 0.83, `sendCustomMessage` with `triggerTurn: false` synchronously does `agent.state.messages.push(appMessage)` and `sessionManager.appendCustomMessageEntry(...)`. `convertToLlm` then maps every `role: "custom"` message to a user message and does **not** honor `display: false` (`content: ""` becomes `[{ type: "text", text: "" }]`). The probe therefore (1) writes a durable session entry and (2) injects an empty user turn into every subsequent prompt/compaction. Several providers reject empty user content; even when they accept it, the host transcript is polluted on every session, including pi, because bindCore capture was removed and session_start now always probes when the steal map is empty. Child `workflow:` / `noExtensions` sessions are still filtered, so this is not a child-leak, but it is a host-context leak.

**Suggestion:** In the prototype wrapper, treat `DELIVERY_PROBE_CUSTOM_TYPE` as capture-only: populate the steal map, then `return Promise.resolve()` without calling `original`. That keeps the void `pi.sendMessage` wrapper thenable-safe and leaves `agent.state.messages` / the JSONL untouched. Restore `_bindExtensionCore` capture as the primary path so pi never needs to probe at all.

src/task-panel.ts
```diff
@@ -209,6 +209,35 @@ const boundSessionSends = new Map<string, DeliverySend>();
...
+ * AgentSession.prototype patch, so no thenable send is ever captured. One
+ * probe through pi.sendMessage forces it: the host's void wrapper calls the
+ * live session's sendCustomMessage synchronously, the prototype patch captures
+ * the receiver, and boundSessionSends is populated by the time this returns.
+ * display:false + triggerTurn:false appends a durable invisible entry on every
+ * supported host and starts no turn.
+ */
+function probeHostSessionSend(pi: ExtensionAPI, sessionId: string): void {
+  if (probedSessionIds.has(sessionId)) return;
+  probedSessionIds.add(sessionId);
```
QuintinShaw: **[bug]** `probedSessionIds.add(sessionId)` runs **before** `sendMessage`, and neither a thrown probe nor a probe that captures the wrong session (last-bindCore-wins `pi.sendMessage`, the exact routing bug this module exists to avoid) clears the set. `dropSessionDelivery` also does not clear it (function itself is unchanged). Quiet omp hosts have no other capture path, so a first-bind miss — including the test’s `"Extension runtime not initialized."` case — makes later `session_start` / `/reload` skip the probe and stay fail-closed. That means this PR can remove the #170 `/reload` workaround when the first probe works, and can make `/reload` **stop** recovering when it does not.

**Suggestion:** Mark a session probed only after `boundSessionSends.has(sessionId)`. If the steal map is still empty, allow the next `bindSessionDelivery` to retry. Clear the id in `dropSessionDelivery`. Combined with the capture-only probe, retries will not append extra transcript rows.

src/task-panel.ts
```diff
@@ -265,18 +307,29 @@ function patchAgentSessionCapture(): void {
...
   agentSessionPatched = true;
   try {
-    const proto = AgentSession.prototype as unknown as {
-      _bindExtensionCore?: (runner: unknown) => unknown;
-    } & StealCandidate;
-    const original = proto._bindExtensionCore;
-    if (typeof original !== "function") return;
-    proto._bindExtensionCore = function patchedBindExtensionCore(this: StealCandidate, runner: unknown) {
+    const proto = AgentSession.prototype as unknown as StealCandidate;
+    const original = proto.sendCustomMessage;
```
QuintinShaw, line 322: **[suggestion]** Capture moved from `AgentSession.prototype._bindExtensionCore` to `sendCustomMessage`. On pi 0.83, `_bindExtensionCore` still runs at session construction (`agent-session.js` calls it before extensions’ `session_start`) and is the session-stable `this` the old steal used — no message needed. Dropping it forces every host, including pi, through the probe / `pi.sendMessage` path. The probe exception around the persistence gate is then load-bearing instead of omp-only.

**Suggestion:** Patch both hooks. Keep the `workflow:` name + `noExtensions` gates on every capture. Use the probe only when `boundSessionSends` is still empty after bindCore. Do not justify the persistence bypass with the name gate alone; unnamed in-memory children rely on `noExtensions` / `isPersisted()`.

src/agent.ts
```diff
@@ -260,49 +260,46 @@ function ensureFallbackRegistry(): Promise<ModelRegistry> {
...
+        const authStorage = (await discover(dir)) as ConstructorParameters<typeof ModelRegistry>[0];
+        return new ModelRegistry(authStorage);
+      }
+      const create = (
+        ModelRegistry as unknown as {
+          create?: (opts: { dir: string }) => Promise<ModelRegistry>;
+        }
+      ).create;
+      if (typeof create === "function") return create({ dir });
+      return undefined;
```
QuintinShaw: **[bug]** `ensureFallbackRegistry` now tries omp’s `discoverAuthStorage`, then a `ModelRegistry.create({ dir })` that this repo’s own previous comment said was **removed in pi >= 0.80.8**, else `return undefined`. That `undefined` is cached (`fallbackRegistry ??= registry`) and asserted as `Promise<ModelRegistry>`. Upstream pi 0.83 (peer `>=0.80.8`, installed `@earendil-works/pi-coding-agent@0.83.0`) has `ModelRuntime.create` + `new ModelRegistry(runtime)`, no `discoverAuthStorage`, and no `ModelRegistry.create`. Unit tests stay green because they inject registries; a `WorkflowAgent` without a host registry that then hits `resolveModelSpecWithThinking` will throw on `modelRegistry.getAll()`, and `listAvailableModels()` stays `[]` forever. This is a regression on the published pi host, not an omp-only no-op.

**Suggestion:** Keep a three-way fallback: `discoverAuthStorage` (omp) → `ModelRegistry.create` (legacy) → `ModelRuntime.create` + `new ModelRegistry(runtime)` (pi >= 0.80.8). Only cache a real registry; do not cache `undefined` as success.

src/agent.ts
```diff
@@ -942,13 +954,25 @@ export class WorkflowAgent {
...
         // spread below re-applies the same injected value harmlessly.
         resourceLoader: this.sessionOptions.resourceLoader ?? (await this.getSharedResourceLoader(agentDir)),
-        // Share the resolved registry's ModelRuntime (catalog + auth, including
-        // extension-registered providers) with the subagent session. pi >= 0.80.8
-        // takes modelRuntime here; the old modelRegistry option is gone.
-        ...(modelRuntime ? { modelRuntime } : {}),
+        // pi >= 0.80.8: createAgentSession takes modelRuntime, not a registry
+        // (the registry option is ignored there). omp's fork has no `.runtime`,
+        // so this spread is a no-op on omp.
+        ...(modelRegistry
```
QuintinShaw: **[suggestion]** Previously the code spread `modelRuntime` only when `runtimeOf()` was truthy. Now any truthy `modelRegistry` spreads `{ modelRuntime: runtimeOf(modelRegistry) }`, which is `{ modelRuntime: undefined }` on omp (no `.runtime` field). The comment says that is a no-op, but it never passes omp’s `modelRegistry` option either, so subagents still do not share the host catalog/auth; the old one-time warning was also removed. On pi, `undefined ?? ModelRuntime.create(...)` happens to rebuild a default runtime, so the host registry handoff still depends on a defined `.runtime`.

**Suggestion:** Restore `const modelRuntime = runtimeOf(modelRegistry); ...(modelRuntime ? { modelRuntime } : { modelRegistry })` (or an equivalent omp-only `modelRegistry` key) so the handoff is explicit on both hosts and `undefined` is never spread.

QuintinShaw (changes requested):
## Summary

The omp delivery probe is a plausible diagnosis (quiet hosts never call `sendCustomMessage`, and a lazy-persist host fails the on-disk gate at `session_start`), and the steal-map still fail-closes when no send is captured. Two correctness problems block merge as-is: the probe is forwarded into the live session, so `convertToLlm` turns it into an empty user message on every later LLM call; and the `ModelRuntime.create` disk fallback is deleted, so stock pi 0.83 (this package’s peer) can no longer build a registry. Issue #170’s “result only shows after `/reload`” is the delivery half this aims to fix; the task-panel widget still hides on complete (running/paused only) and is unchanged.

## Issue counts by severity

- bugs: 3
- suggestions: 2
- nits: 0

src/task-panel.ts
donrami: Fixed in b554303. `patchedSendCustomMessage` now treats `DELIVERY_PROBE_CUSTOM_TYPE` as **capture-only**: it calls `captureHostSessionSend(this, true)` and returns `Promise.resolve()` without invoking the original send — nothing is appended, persisted, or turned.

- Regression test: `"probe is capture-only: never forwarded to the session sendCustomMessage"` — asserts `agent.state.messages` and `sessionManager.appendCustomMessageEntry` see nothing for a probe.
- Live proof on both hosts: after the E2E runs, session transcripts contain **zero** `workflow-delivery-probe` entries (previously one durable entry per session on omp).

src/task-panel.ts
donrami: Fixed in b554303. The mark now happens **after** a send is actually captured, in both failure modes:

```ts
} catch (err) {
  // Not marked probed — the next bindSessionDelivery retries.
  return;
}
if (boundSessionSends.has(sessionId)) probedSessionIds.add(sessionId);
```

- A thrown probe returns without marking; a probe that captures nothing (wrong routing, child-session rejection) fails the `has` check — both retry on the next bind.
- `dropSessionDelivery` also deletes the mark, so rebind-after-replace probes fresh.
- Tests: `"failed probe is retried on the next bind (no pre-marked probed set)"` and `"dropSessionDelivery clears probed state: a new bind probes again"`.

Writing the retry test exposed one more latent bug: `deliverAndAck` took the in-flight lock even on the fail-closed (no-send) path, so a synchronous re-bind + flush was locked out by an un-run microtask. The lock is now taken only when a real endpoint send exists.

src/task-panel.ts
donrami: Kept both hooks. The construction-time capture via `_bindExtensionCore` is retained (`patchBindCoreObserve` still armed, its patched forward captures with `isProbe=false`), and the receiver-scoped capture in `patchedSendCustomMessage` covers hosts where the session id is not final until after bindCore (omp persists the host session lazily, so `isSessionOnDisk` is false at construction but the later receiver capture sees the settled id).

They are idempotent rather than competing: `captureHostSessionSend` is keyed by sessionId and the full guard chain (noExtensions, `workflow:`-named child rejection, persistence verdict, probe-vs-real distinction) runs on every capture, so whichever fires first wins and the second is a no-op for the same id.

Regression test for the construction path: `"bindCore capture path captures a persisted host without any probe"` — a persisted host is stolen with zero `pi.sendMessage` calls.

src/agent.ts
donrami: Fixed in b554303. The chain now ends in a **real ModelRegistry** on every host — no `undefined` return:

1. omp fork exports `discoverAuthStorage` → `new ModelRegistry(await discoverAuthStorage(dir))`.
2. Legacy `ModelRegistry.create({ dir })` — verified present in pi 0.84.3 dist (static `create` on `ModelRegistry`), so this branch is live on stock pi.
3. `ModelRuntime.create(...)` wrapped in `new ModelRegistry(runtime)`.

Rejections are never cached: `fallbackRuntimePromise.catch(() => { fallbackRuntimePromise = undefined; })` so a transient `auth.json` lock cannot wedge the process.

Tests pin the contract: `"fallback registry resolves to a real ModelRegistry on stock pi and is cached across agents (never undefined)"` (runs under a fake HOME on real stock pi types), plus `"stock pi does not export discoverAuthStorage"` so branch 1 can never mask 2/3 on upstream.

src/agent.ts
donrami: Fixed exactly as suggested in b554303:

```ts
const modelRuntime = runtimeOf(modelRegistry) as ModelRuntime | undefined;
...
...(modelRuntime ? { modelRuntime } : { modelRegistry }),
```

`runtimeOf` is computed once; only a **defined** runtime is spread. On stock pi the registry key keeps the host catalog/auth handoff explicit when the runtime field is absent; on omp the fork (which takes `modelRegistry` after 0.80.8 dropped it) resolves through the same ternary. The cast is only needed because upstream `CreateAgentSessionOptions` has `modelRuntime` but not `modelRegistry` — the runtime value itself is unmodified.

donrami:
All five review findings are addressed on this branch (rebased onto current `main`, head `b554303`) — replies with specifics are on each thread:

1. **Probe pollution** → capture-only wrapper; probe never reaches the real send. Transcripts on both hosts now contain zero probe entries.
2. **Probe retry** → probed-set marked only after capture; failed/wrong probes retry on the next bind; `dropSessionDelivery` clears the mark. (Also fixed a latent in-flight-lock bug the retry test surfaced.)
3. **Capture hooks** → both retained: `_bindExtensionCore` for construction-time capture on pi, receiver-scoped `sendCustomMessage` capture for post-bind ids; idempotent per sessionId.
4. **Fallback registry** → 3-way chain ends in a real `ModelRegistry` on every host, rejection never cached; verified against pi 0.84.3 dist shapes.
5. **`modelRuntime` spread** → computed once, never spread when undefined; registry key keeps the handoff explicit on both hosts.

Verification beyond the unit suite (1312 tests, biome + tsc clean; the two failing tests on `2f85c9e` — `workflow-authoring-skill`, `workflow-release-gate` — fail identically on pristine upstream, unrelated to this PR):

- **Live omp E2E**: background run delivered a `workflow-result` turn, `pendingDelivery` cleared to `null`, no probe entry persisted.
- **Live stock pi 0.84.3 E2E** (extension mounted via `-e`, separate auth): same delivery, same zero-probe transcript. This also exercises the retry path — pi had no steal-map entry at all, so capture came entirely from the probe.

QuintinShaw (approved):
Re-reviewed `b554303` against the previous CHANGES_REQUESTED items.

All five are fixed in source:

1. Probe is capture-only. `patchedSendCustomMessage` swallows `workflow-delivery-probe` and never calls the original send, so nothing is appended, persisted, or turned into an LLM user message.
2. `probedSessionIds` is marked only after a send is actually captured; a thrown/missed probe stays unmarked and is retried. `dropSessionDelivery` clears the mark.
3. Construction-time `_bindExtensionCore` capture is kept; the sendCustomMessage patch remains the quiet-host / probe fallback.
4. Fallback registry construction is a real three-way path (`discoverAuthStorage` → `ModelRegistry.create` → `ModelRuntime.create`) and throws if none exist — stock pi 0.83 reaches `ModelRuntime.create`.
5. `createAgentSession` spreads `{ modelRuntime }` only when the runtime is defined; otherwise `{ modelRegistry }`. Fail-closed binds no longer take the in-flight lock.

`test` is green on this head.

Please drop the unrelated noise before merge (`package-lock.json` metadata, `model-spec.ts` / `workflow-comprehension.ts` / `workflow-editor.ts` / the `workflow-release-gate.ts` comment). Those are not part of the delivery fix.

Not blocking this PR: the generation-change retry in `deliverAndAck` still deletes the in-flight lock in `.then` and again in `.finally`, which is the same dance as on `main` today. Worth a follow-up so a failed send overlapping a re-bind cannot double-deliver, but it is not introduced here.

donrami:
Follow-ups from your review: the double-delivery race you flagged in `deliverAndAck` is #175, fixed by #176 (ownership-token in-flight lock + regression test, verified on v3.10.0 — 1331/1333 tests, the 2 failures pre-exist on pristine `main`; live omp E2E delivery green).

On the pre-merge "drop the unrelated noise" request — understood for next time; the lockfile/model-spec/comprehension hunks rode along in the squash.

Holding off on the reversion PR unless you still want it.

PR #176, opened 2026-08-30
fix(delivery): ownership-token in-flight lock prevents stale finally double-delivery

### LINKED ISSUE #175: Delivery: a failed send overlapping a re-bind can double-deliver workflow results
Found while re-reading `deliverAndAck` in `src/task-panel.ts` (as shipped in #169).

**Race.** When a send fails and the endpoint generation changed meanwhile, the `.then` handler releases the in-flight lock, flushes disk-pending, and the flush immediately re-acquires a **new** lock for the retry send. The original promise chain's `.finally` then runs — and unconditionally deletes the runId from `inFlightDeliveries` — releasing the **retry** send's lock while it is still in flight. A third caller (another bind flush or a duplicate `complete` event) can now pass the `has()` guard and start a second `workflow-result` turn for the same run.

Sequence:
1. Send A (gen N) in flight, lock held.
2. Gen N+1 binds. Send A rejects.
3. `.then`: delete lock → flush → retry send B starts, lock re-acquired.
4. `.finally` (from A's chain): delete lock again → B unprotected.
5. Another flush/complete for the same runId → duplicate delivery turn.

**Fix.** Make the lock ownership-tokened: `Map<runId, token>`; a chain releases only if its token is still the owner. Regression test included in the follow-up PR.

Found by the maintainer's review follow-up on #169; fix incoming.

donrami:
Fixes #175.

`deliverAndAck`'s in-flight lock (`Set<runId>`) could be released twice for one run: the failed send's `.then` releases it, the generation-change retry re-acquires it for the next send, and the original `.finally` deletes it again — leaving the retry send unprotected so a third caller (bind flush / duplicate `complete`) can double-deliver the same `workflow-result` turn.

**Change:** the lock becomes `Map<runId, token>`. Each deliver-and-ack captures its own token at acquisition; both `.then` and `.finally` release via `release()`, which deletes only if the stored token still matches. A superseded chain can no longer drop a newer send's lock.

**Verification:**
- New regression test `a failed send overlapping a re-bind cannot double-deliver` — fails on the old `Set` lock (retry released mid-flight), passes with the token lock.
- Full suite: 1331/1333 — the 2 failures (`workflow-authoring-skill`, `workflow-release-gate`) fail on pristine `main` too (pack-shape/environment, pre-existing).
- biome + tsc clean; live omp E2E on this tree: workflow launched in background, `workflow-result` delivered with `pendingDelivery` cleared, zero probe entries.

### DIFF
--- a/src/task-panel.ts
+++ b/src/task-panel.ts
@@ -206,8 +206,9 @@ const sessionEndpoints = new Map<string, SessionDeliveryEndpoint>();
  */
 const boundSessionSends = new Map<string, DeliverySend>();
 
-/** runIds with an in-flight deliver-and-ack so bind flush does not double-send. */
-const inFlightDeliveries = new Set<string>();
+/** runId → token of the deliver-and-ack that owns the lock, so bind flush does not double-send. Ownership prevents a stale finally from releasing a newer send's lock. */
+const inFlightDeliveries = new Map<string, number>();
+let inFlightSeq = 0;
 
 /**
  * Custom type for the session bind probe (see probeHostSessionSend). Sent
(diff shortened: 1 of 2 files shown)


QuintinShaw (approved):
Thanks — this is the right fix for the stale-finally double-release. Ownership-tokened Map so a superseded chain cannot drop a newer send's lock, and the regression test fails on the old Set.

Approving.

PR #177, opened 2026-08-31
fix(delivery): make background workflow results reliable

### LINKED ISSUE #175: Delivery: a failed send overlapping a re-bind can double-deliver workflow results

Ancientkingg:
Background workflow results can remain pending after a transient send failure, or be acknowledged while still queued in memory. Session replacement can then discard that queue.

This change:
- assigns each pending result a stable delivery ID and records it in the custom message
- waits for matching session persistence before acknowledging streaming follow-ups
- retries transient same-session failures with bounded backoff
- prevents stale send completions from releasing newer locks or clearing newer markers
- retries marker clears and keeps undelivered terminal runs out of retention eviction

It includes the ownership-token lock fix from #176, plus coverage for rejected sends, session replacement, persistence failures, stale acknowledgments, and restart deduplication.

Fixes #175.

`npm test` passes.


QuintinShaw (changes requested):
Thanks — the ownership-token lock, persist-before-send, delivery-id reuse, failed-clear retry, and undelivered retention are the right shape, and the new tests pin #175 and replacement resend.

I can't merge this head as-is. The streaming ACK waiter waits forever for a matching `message_end` and holds the in-flight lock the whole time:

- `sendCustomMessage({ deliverAs: "followUp" })` resolves after queueing (`agent.followUp`)
- the wrapper then waits for `message_end` with no timeout, no `agent_end`/`abort` observer, and no fail if the follow-up is discarded
- `Promise.resolve(queued).catch(finish)` cannot recover because that promise already resolved
- same-session retry (`scheduleDeliveryRetry`) never runs while `inFlightDeliveries.has(runId)`
- session replacement recovers only because `suspendSessionDelivery` deletes the lock; interactive Escape goes through `restoreQueuedMessagesToEditor` → `session.clearQueue()` → `agent.clearAllQueues()` and does not suspend

That leaves a same-session abort able to pin the result as `pendingDelivery` until the next bind — the opposite of the reliability claim.

Please fail the waiter when the follow-up cannot still land (timeout and/or `agent_end`/`queue_update` + `sessionHasDelivery`), so the existing same-generation retry can send on the idle path. Also unsubscribe/reject on suspend so a zombie waiter cannot ACK a later generation that reused the same `deliveryId`.

Non-blocking after that:
- ACK success should require `current.generation === startedGeneration` (or the live endpoint identity), not just session id
- the streaming persist test should fire `message_end` before appending history, matching AgentSession

#176 is landing separately for the lock-only #175 fix so this PR can rebase onto that.

Ancientkingg:
Rebased onto main and addressed the feedback:

- The streaming waiter now rejects and releases the in-flight lock on `agent_end`, `agent_settled`, queue clearance (`queue_update`), or timeout when the message has not been persisted. That lets same-generation retry deliver on the idle path.
- `suspendSessionDelivery` now cancels and unsubscribes active streaming waiters so old waiters cannot ACK later generations reusing the delivery ID.
- Delivery ACKs now require `current.generation === startedGeneration` and live endpoint identity.
- Adjusted the streaming persistence test to fire `message_end` before appending history.
- Added tests covering same-session aborts, queue clears, timeouts, suspend cancellation, and stale generation ACKs.

QuintinShaw (changes requested):
Thanks for adding waiter cancellation, generation checks, and the new regression cases. The latest CI passes, but the streaming recovery still has a duplicate-delivery path.

1. An ACK timeout does not mean the queued follow-up was discarded. The timeout rejects the waiter, releases the lock, and schedules a retry after 250 ms. If the host is still streaming, that retry queues the same delivery ID again. The first message is still in the agent's follow-up queue, and `sessionHasDelivery` cannot deduplicate either copy before history contains it. A long-running turn can therefore queue the result repeatedly. Please preserve delivery ownership while the original message may still arrive, or establish that the original was removed before enqueueing another copy. Cover a turn that remains streaming beyond the ACK deadline, then drains its queue.

2. `queue_update.followUp` is not the custom-message queue. In the installed Pi SDK, this event reports `_followUpMessages`, the user-text queue, while `sendCustomMessage(..., { deliverAs: "followUp" })` directly calls `agent.followUp(appMessage)`. An unrelated steering update can therefore report `followUp: []` while the workflow result is still queued. The current OR condition treats that as a discard even when `agent.hasQueuedMessages()` is true, producing the same duplicate retry. Please test a steering update with a workflow result still queued and avoid using an empty user-text queue as proof that the custom message was removed.

Both cases must deliver the result exactly once after the original queue drains. The existing timeout test switches the host to idle before retry, so it does not exercise the first case.

QuintinShaw:
Superseded by #193, now merged on main. It retains the durable pending-marker/retention work and replaces ambiguous streaming queue acknowledgements with current-branch session-file confirmation and explicit send ownership. Real Pi SDK, model-backed lifecycle tests, and interactive-terminal delivery passed. Thank you for the contribution; no package release has been published yet.

PR #178, opened 2026-08-31
perf(extension): load the compiled entrypoint directly

gurronen:
## Summary
- avoid transpiling the workflow extension's TypeScript graph during every Pi startup

## Changes
- compile the Pi extension entrypoint with the existing TypeScript build
- point `pi.extensions` at `dist/pi-extension.js`
- replace the extension's package-barrel import with direct module imports
- verify the compiled entrypoint is included in the npm package and exports an extension factory

## Testing
- [x] `npm test`
- [x] `npm pack --dry-run --json`
- [x] manual `pi -e .`
- [x] startup benchmark before/after

## Startup benchmark
Measured in a fresh Bubblewrap sandbox per run, with networking disabled, an empty temporary home/cache, all normal Pi extension/skill/theme/context loading disabled, and only this extension explicitly enabled. One warm-up was discarded and 10 alternating runs were recorded.

| Case | Entrypoint | Median | Range |
| --- | --- | ---: | ---: |
| Without PR (`4aaf451`) | `extensions/workflow.ts` | 1.941s | 1.505–2.245s |
| With PR (`f3a3de9`) | `dist/pi-extension.js` | 0.949s | 0.784–1.094s |
| Pi control, no extension | — | 0.234s | 0.220–0.261s |

The compiled entrypoint reduced median total startup by **0.991s (51.1%)**, or **2.04×**. Relative to the no-extension control, median extension overhead fell from approximately 1.707s to 0.715s (58.1%).


QuintinShaw (approved):
Thanks — compiling the extension entry and pointing `pi.extensions` at `dist/pi-extension.js` is the right way to stop paying a TypeScript transpile on every Pi startup. The factory move is faithful, pack-shape is tested, and npm consumers get `dist/` via `files` + `prepublishOnly`.

One non-blocking nit: a fresh clone or `pi -e .` before `npm run build` silently loads nothing, because Pi skips missing manifest paths. Worth a Development-section note that local `pi -e .` needs `npm run build` first. Optional later: a tiny committed trampoline that fails loud if dist is missing. Do not add a TS fallback next to the compiled path.

Approving.

PR #179, opened 2026-09-02
feat: add pre-spawn model resolver hook

fishfuuu:
## What

External extensions currently cannot deterministically customize or reject the final worker model after Dynamic Workflows resolves model/tier intent but before `createAgentSession`.

This PR adds an **opt-in async pre-spawn resolver**:

- `setPreSpawnModelResolver` / `getPreSpawnModelResolver` (public package exports)
- Decisions: `unchanged` | `use` + concrete model | `reject` + reason
- `modelSource`: `explicit` | `tier` | `phase` | `default` | `session`
- Per-run `preSpawnModel` on `WorkflowAgent.run` overrides the process resolver (tests/embedders)

## Why

Hosts and independent Pi extensions need a composition point for:

- organization policy
- availability policy
- custom model selection

This is not a quota-specific router, not a middleware framework, and not a change to default routing.

## Backward compatibility

No resolver registered = existing v3.10.0 routing (explicit, tier, phase/default, session fallback).

## Intentional limitation

Single process-wide resolver. Last `setPreSpawnModelResolver` call wins. Stored on `globalThis` under `Symbol.for("@quintinshaw/pi-dynamic-workflows.preSpawnModelResolver")` so an independent extension and the packaged/dist runtime share one slot. No middleware chain, priority, or registry. The resolver is not serialized into worker context.

## Validation

- `tsc` build and `tsc -p tsconfig.scripts.json`
- Focused tests in `tests/pre-spawn-model.test.ts` (13/13)
- Isolated Pi environment: external extension registers via the public import
  - C1: `tier: 'small'` → automatic override, actual worker session on the override model
  - C2: reject → `MODEL_SPAWN_REJECTED`, no successful worker session
  - C3: no resolver → existing `MODEL_NOT_FOUND` for an unavailable configured small tier

On Windows with `core.autocrlf=true`, full-repo `biome check .` / `docs:check` fail identically on unmodified v3.10.0 (CRLF). CI runs on Ubuntu.


QuintinShaw (changes requested):
Thanks — the seam is the right size (one opt-in policy after DW resolves intent, before createAgentSession), reject is fail-closed, and nested workflow() cannot miss the in-process resolver. Process-global last-write-wins on `Symbol.for` is acceptable interop for independent extensions, not a middleware chain.

Please address these before merge:

1. Diagnostic bug: `action: "use"` forces `pinAfterPolicy = true`, so a missing override throws `MODEL_NOT_FOUND`. The message still uses `options.model ? … : tier "${options.tier}"`. A session/default `use` therefore reports `tier "undefined" from model-tiers.json`. If the decision was `use`, name the policy spec; do not mention tier unless that spec actually came from tier resolution. Cover session and tier `use` → missing model in tests.

2. Public `ModelSource`: `"default"` currently means untagged medium-tier **or** phase/meta routing. The README sample remaps every non-explicit spawn, including user `tier` pins and author phase routes, and DW will not log a silent remap of `explicit`. Before freezing the enum, either add `"phase"` (or equivalent) or document the lump as the contract. Change the sample so it does not silently remap `tier` unless that is the intended host pattern. Mention instance-level `WorkflowAgentOptions.preSpawnModel` in the same README paragraph as run/process precedence.

Nits: unknown `action` strings fail-open (whitelist and throw); add CI coverage for process-level `use` on tier, process-level reject, and run > instance > process. The isolated-Pi C1/C2 notes in the PR body are not a substitute for those unit cases.

fishfuuu:
Addressed review feedback in e7fba5c:

- fixed policy-selected missing-model diagnostics
- clarified ModelSource provenance (`phase` vs untagged `default`)
- hardened invalid action handling
- added requested precedence/process-level unit coverage
- updated README example/precedence docs

src/agent.ts
```diff
@@ -931,11 +977,13 @@ export class WorkflowAgent {
...
-          // branch swaps in its own message so the config source is named too.
-          const message = options.model
-            ? (resolved.error ?? `Model "${modelSpec}" not found. Use /workflows-models to choose an available model.`)
-            : `tier "${options.tier}" from model-tiers.json resolves to "${modelSpec}", which is not available. Use /workflows-models to choose an available model.`;
+          // Policy `use` pins the spec: name that spec, not the original tier/model.
+          // Otherwise a session/default `use` would report `tier "undefined"`.
+          const message = policySelectedSpec
+            ? `Model "${modelSpec}" selected by preSpawnModel policy was not found. Use /workflows-models to choose an available model.`
+            : options.model
+              ? (resolved.error ?? `Model "${modelSpec}" not found. Use /workflows-models to choose an available model.`)
```
QuintinShaw: CI failed here: Biome wants this ternary wrapped.

```suggestion
              ? (resolved.error ??
                `Model "${modelSpec}" not found. Use /workflows-models to choose an available model.`)
```

QuintinShaw (changes requested):
Thanks — the previous review items are actually fixed: policy `use` names the policy spec (T4/T5), `ModelSource` has `phase`, the README sample no longer remaps `tier`/`explicit`/`phase`, instance precedence is documented, unknown actions fail closed, and process-level use/reject plus run > instance > process are covered.

CI is red on a biome format wrap in `src/agent.ts` (`npm test` → `biome check`). I could not push to the fork from here. Please apply the inline suggestion (or `npx biome check --write src/agent.ts` with this repo's Biome) and I'll re-run CI.

QuintinShaw (commented):
The previous code-review requests and the formatting fix are resolved. The branch now includes current main, and CI passes on 725b6d5. The public package setter and source runtime share the resolver slot; phase provenance and reject-before-spawn checks also pass. I updated the PR description's ModelSource list to include `phase`.

The remaining merge gate is the real Pi default/use smoke required by CONTRIBUTING for model-routing changes. The available local provider configurations rejected those calls before a successful response, so that validation is incomplete. This is a validation-environment blocker, not a new code-change request. I am keeping the PR unmerged until default routing and a policy-selected model both complete a real subagent turn.

fishfuuu:
Completed the remaining real-Pi E2E smoke gate on PR head `725b6d5`.

### Default / no resolver — PASS

- Resolver explicitly cleared
- Actual model: `ollama/deepseek-v4-flash:0731`
- Real `WorkflowAgent.run -> createAgentSession -> real model` turn completed
- `onModelResolved`: `ollama/deepseek-v4-flash:0731`
- Terminal response: `DEFAULT_SMOKE_OK`

### Policy `action:"use"` — PASS

- Requested/resolved model before policy: `ollama/deepseek-v4-flash:0731`
- `modelSource`: `explicit`
- Policy-selected model: `ollama-copy/glm-5.3-flash`
- Actual spawned model: `ollama-copy/glm-5.3-flash`
- The spawned model differs from the requested model and matches the policy selection
- Real subagent turn completed
- Terminal response: `POLICY_SMOKE_OK`

Validation used a temporary repo-root harness importing the current PR implementation from `./src`.

Cleanup:
- resolver cleared after validation
- temporary harness deleted
- working tree clean
- HEAD unchanged at `725b6d5`
- no commit or push

QuintinShaw (approved):
Rechecked the updated head `a82f16edc6c30e08875e604ced36a4d9fbdbeba5` against Pi SDK 0.85.1. The 21 resolver tests and build pass.

Real `WorkflowAgent -> createAgentSession -> openai-codex/gpt-5.6-luna` verification now passes independently:

- With no resolver installed, the requested low-thinking model returns `DEFAULT_OK`.
- A policy receives an explicit high-thinking request and selects the low-thinking variant. The resolver context retains the original request, `onModelResolved` reports the selected variant, and the real call returns `USE_OK`.
- A rejected spawn raises non-recoverable `MODEL_SPAWN_REJECTED` without a model-resolution callback or provider call.
- The public compiled setter and source runtime share the same resolver slot.

The previous real-provider verification blocker is cleared. No remaining code findings in this change; merge remains conditional on green checks for the updated head.

PR #180, opened 2026-09-06
feat: keep isolation worktrees by default; isolation: false opt-out

kaankoken:
## Why

`isolation: "worktree"` always tore down the tree in a `finally` block (force-remove the worktree and drop its branch). Implementer agents that wrote in that tree lost the work — never merged, never on disk.

There was also no way to opt out of an `agentType` isolation default (`isolation: undefined` fell through to the def).

## What

- **Keep by default.** After an isolated call, log `worktree kept: <path> (<branch>)` so the caller can merge. Not auto-merged.
- **`keepWorktree: false`** deletes after the call (test / ephemeral runs).
- **`isolation: false`** opts out of an agentType worktree default. Isolation stays optional at the call site.
- Threads still cannot combine with worktree isolation.

## Tests

`npm test` green (1336). New cases in `tests/agent-registry.test.ts`: default keep, `isolation: false` opt-out, `keepWorktree: false` teardown.

**Note:** default keep is a behavior change vs always-delete. Callers that relied on auto-cleanup should pass `keepWorktree: false`.


mindplay-dk:
makes sense, but will the worktrees get cleaned up at all? after you close pi?

or will they potentially pile up after aborted or failed runs etc.?

QuintinShaw (changes requested):
Thanks for preserving isolated agents' work. There is a resume regression to address before this can merge.

`createWorktree` derives the same branch and path from the run and call identity. With the new keep-by-default behavior, a second execution of that call finds the retained branch, `git worktree add -b` fails, and the helper returns `isolated: false` with the shared project directory. `runWorkflow` only logs that isolation was ignored and then runs the agent there. I reproduced this with two calls to `createWorktree` using the same name in a real temporary Git repository: the first was isolated; the second fell back to the repository root because the branch already existed. An interrupted or failed call that executes again on resume can therefore silently lose its requested isolation.

Please define and test how retained worktrees are recovered or replaced on resume. Failure to obtain the requested isolation must not silently move edits into the shared checkout. Include success, failure, abort, and repeated-resume coverage with a real Git repository.

On the cleanup question above: this patch keeps the tree in `finally` on success, failure, and abort unless `keepWorktree: false` is passed. It does not add a session-exit cleanup path. Please document who owns those retained trees, how users discover and clean them up, and how unmerged work is protected. The default change also needs an explicit compatibility policy.

Finally, please move the separate `thinking` configuration feature into its own PR, as requested by CONTRIBUTING's one-concern-per-PR guidance. That change needs its own routing, precedence, and replay review.

QuintinShaw:
Superseded by two focused PRs, both now merged: #190 carries the worktree retention/opt-out behavior with fail-closed creation and independent ownership; #194 carries the separate thinking option with policy/precedence and resume-identity coverage. Both passed real Pi validation and CI. Thank you for the contribution. No package release has been published yet.

PR #182, opened 2026-09-07
feat(runtime): warn when a workflow's agents all return null (empty fleet)

yaokuku123:
## Why

`agent()` deliberately resolves an exhausted **recoverable** failure (e.g. `AGENT_EMPTY_OUTPUT` after retries) to `null` instead of throwing, so one bad agent never sinks the whole fleet — and `parallel()`/`pipeline()` resolve with `Array<unknown | null>`. That contract is documented and good.

But there is an observability gap at the fleet level: when **every** agent comes back `null`, the run still reports `completed` and the only signal is a single per-agent log line, which is easy to miss under concurrency. The orchestrating model/host can mistake "nothing was produced" for "everything succeeded."

This bit me for real: a 5-agent fan-out reported `ok` with every report `null`. The underlying cause in my case was a model whose output-token cap made heavy first-turn reasoning truncate to an empty response — easy to miss when the run looks green. A one-line safety net would have surfaced it immediately.

## What this PR does

At run completion, when **at least one agent ran and every terminal agent returned null**, the runtime emits one prominent warning (before the `complete` event):

```
⚠  Workflow produced no usable results: all 3 agent(s) returned nothing (a, b, c).
   agent() resolves recoverable failures (e.g. an empty model response) to null once retries are exhausted, so the run reports "completed" even though nothing was produced.
   Common fixes: set agentRetries: 1-2 for models that occasionally return empty, check the model's max output tokens, and read the per-agent errors above.
```

It goes through the same `onLog` channel as other runtime logs (so it lands in the run logs / persisted `logs` and the live `"log"` event) **and** stderr. Empty agents are named (capped at 5, with `+N more`).

**The null / recoverable contract is unchanged** — the run still completes, `agent()` still returns `null`, no retry semantics change. This is observability only; nothing is gated or failed.

## Changes

- `src/display.ts`: `emptyFleetSummary()` pure helper + `EmptyFleetSummary` type. Counts terminal `error` (null result) vs `done` (real result); ignores `queued`/`running`/`skipped` so nothing cries wolf mid-run or on no-op/skipped fleets.
- `src/workflow-manager.ts`: at run completion, if `allEmpty`, push the warning through the existing `onLog` wiring + `console.warn`.
- `README.md`: documents the warning next to the existing `AGENT_EMPTY_OUTPUT` paragraph.
- `tests/empty-fleet.test.ts`: 6 unit tests (all-error → flagged; one success → not; no agents → not; still-running → not; only-skipped → not; label capping + unlabeled fallback).

## Tests / verification

- `npm test` green: Biome, `tsc` build, docs/context/guidance checks, and all **1340** unit tests pass (including the 6 new ones); release gate passes with 0 warnings.
- Per CONTRIBUTING’s "runtime behavior" guidance, I verified end-to-end against a real `WorkflowManager` + fake agent runner (throwaway harness in repo root, deleted before commit):
  - empty-whitespace agent ×3 → `AGENT_EMPTY_OUTPUT` → null → warning appears in both the live `"log"` event and the persisted run logs;
  - a healthy fleet → **no** warning (no false positive).
- Not a change to agent execution/retries/routing/concurrency, so no real-provider model session was required for the control flow; the warning path itself was driven through the actual manager completion flow rather than a mock.

Happy to adjust wording, threshold (e.g. warn at >x% null rather than 100%), or route the warning differently.


QuintinShaw (approved):
Reviewed the warning against the manager's result-to-status mapping and completion path. All-null results emit one warning in the live log event and both in-memory and persisted logs. A mixed successful/null run and a no-agent run do not warn; nested calls and cached null replay follow the existing status mapping. The completion and recoverable-null semantics remain unchanged.

The focused tests pass, the manager integration scenarios above were exercised locally, and CI passes on 5dc200c. Approved.

Issue #170, opened 2026-08-28
Workflows not Updating Session Status


chris-fenton-rg:
Since yesterday, I've seen multiple Pi sessions run into issues where the workflows plugin wasn't updating the main session when it finished. The bottom widget would disappear, and I would have to run `/reload` which would show the workflow result tool output and the session would start moving again. Some other sessions didn't have this problem.

QuintinShaw:
Thanks for the report. There are two separate things in what you described:

1. **Result only appears after `/reload`** — that's conversation delivery. Background completion is fail-closed: if this process never captured a session-stable send, the result stays `pendingDelivery` on disk and only flushes when a later bind (exactly what `/reload` does) captures one. Quiet hosts that never call `sendCustomMessage` themselves never trip that capture until reload.

2. **Widget disappearing when the run finishes** — the task panel only lists `running` / `paused`. A completed run dropping out of the overlay is existing behavior, not a stuck session.

Could you add:

- Host (`pi` CLI vs omp / another fork) and `@quintinshaw/pi-dynamic-workflows` version
- Whether the run JSON still has `pendingDelivery` after the widget vanishes
- Whether `/reload` prints a `workflow-result` custom message, or only unsticks the session

There's an in-flight attempt to fix (1) for lazy-persist hosts (#169). I'm not pointing people at it yet — the current patch has correctness issues that need another round.

Related: #171 covers a different abort/fail persistence bug (agents left `running` on disk). If your stuck sessions were aborted/failed rather than completed, that record inconsistency is tracked there.

donrami:
Related context from the #169 work: background-run delivery itself (send-steal/probe, ack, disk-pending flush) was reworked in #169 and now holds an ownership-tokened in-flight lock (double-release race fixed in #175/#176). If this symptom involves the run panel not reflecting delivered runs, the delivery-side state transitions it can rely on are: `pendingDelivery` set at completion (persisted), cleared only after a thenable send resolves, and endpoint generation bumps on re-bind. Happy to help dig into the widget half — it looks orthogonal to the delivery lock.

chris-fenton-rg:
Gathered the requested info from the affected session. The session forked before the run finished. This produced a live reproduction of the bug.

**Host and version**

- Host: stock pi CLI 0.84.4 (`/opt/homebrew/bin/pi`). Not omp.
[...]

PR #183, opened 2026-09-08
fix(delivery): recover pending workflow results on session turn_end

Ancientkingg:
Background workflow results can remain pending if delivery finishes while the session endpoint is missing, suspended, or lacking a send function. Because ordinary conversation turns in an existing session do not trigger session_start, those results can stay un-delivered while the user continues chatting.

This change:
- Re-probes missing senders and flushes pending deliveries on turn_end once the session settles.
- Displays completed workflows with pending deliveries in the task panel.
- Logs specific deferral reasons in deliverAndAck when delivery cannot proceed.
- Adds regression tests for turn_end recovery and task panel pending display.

Follow-up to #177.

npm test passes.


QuintinShaw (changes requested):
The recovery hook and pending-result visibility are useful. This head includes #177, so it also includes the streaming duplicate-delivery paths described in that review. Please update this PR after those are fixed; passing CI on the current stack is not sufficient to merge it.

There is also a status-label issue in both panel renderers: every terminal run with `pendingDelivery` is shown as "Completed, result delivery pending", including failed and aborted runs whose pending marker contains the failure text. Please render the actual terminal status and add failed/aborted pending cases alongside the completed case.

Keep #170 open until the revised stack has been exercised in a fresh Pi session, including idle completion and session fork/rebind. This patch does not by itself establish that those reported cases are resolved.

QuintinShaw:
Superseded by #193, now merged on main. The turn-end recovery path is retained, while idle delivery, durable acknowledgements, reload/owner handoff, and accurate pending status are covered by the integrated fix. Real Pi and interactive-terminal validation passed. No package release has been published yet.

PR #188, opened 2026-09-12
feat: Customisable agent cache TTL

johnwards:
Anthropic prices a cache write by the retention requested: the 1h tier costs more per write than the 5m one. A long-lived parent conversation earns that back by surviving idle gaps, but agent sessions are short-lived and rarely idle long enough to claim the longer window, so a wide fan-out pays the higher write price with no benefit.

PI_CACHE_RETENTION=long PI_WORKFLOW_AGENT_CACHE_RETENTION=short keeps the longer window for the parent conversation only, the split Claude Code makes between its main conversation and its workflows.

Unset by default, so agent sessions inherit the parent's retention and behaviour is unchanged unless opted into. 

Documented in README.md alongside the other storage and session settings. npm test passes in full.

Verified end-to-end against a real createAgentSession run on claude-opus-5 with persistAgentSessions, parent and agent on the same model, reading provider-reported cacheWrite1h from the persisted transcripts:

  parent, PI_CACHE_RETENTION=long   cacheWrite 34713  cacheWrite1h 34713
  agent, override short             cacheWrite 10083  cacheWrite1h     0
  agent, override unset             cacheWrite 10083  cacheWrite1h 10083


QuintinShaw (approved):
Reviewed the request-scoped wrapper and the added regression coverage. The unset case preserves the original stream function; configured retention only overlays the child request environment, without mutating the caller options or process environment. Other stream options remain intact.

Validated on Pi 0.85.1: 1,366 tests passed, the real session wiring test observed the override, and actual model requests succeeded both unset and with `short`. This does not claim independent verification of Anthropic cache pricing or expiry. CI is green on 3a68a541236eb62e8ec5eaec76c3a984c7d9c5b9.

PR #198, opened 2026-09-14
feat: link workflow runs to parent and child sessions

codesoda:
## Why

`persistAgentSessions` (#54) preserves full child transcripts, but it does not link those sessions back to the host session or record child session identities in workflow run state. This makes it difficult to trace a workflow call to its transcript or reconstruct parent/child session relationships after a pause or restart.

## Changes

- Capture the host session ID/file when a run starts, independently of its current delivery owner.
- Set Pi's existing `parentSession` header to the originating host session file for persisted children. No parent messages are inherited.
- Record each call's child session ID/file in run persistence and snapshots, keyed by `callId`, not display label.
- Preserve child identities, historical timestamps, and usage when completed calls replay on resume; a replay is not a new child launch.
- Keep original parent lineage when delivery ownership moves to a replacement session.
- Keep in-memory sessions as the default. Ephemeral children have IDs but no transcript file; legacy run files remain readable. No new persistence setting is introduced.

## Verification

- `npm ci`
- `npm test`: all 1,431 tests passed, including build, documentation/context checks, and release verification.
- `npm run guidance:check`
- `git diff --check`
- Added real `SessionManager` header tests and real SDK session tests using a faux provider, plus persistence/replay tests covering duplicate labels, session replacement, and ephemeral children.

**Validation limitation:** live-provider end-to-end validation has not been run. The SDK-path tests use a faux provider; a real-provider smoke test is still needed before merge per CONTRIBUTING.md.

This is an independent PR based on upstream `main`; it does not include the separate settings or extension-allowlist changes.


QuintinShaw:
Thanks for the contribution. Merged after review and green CI on the branch updated with main. Real subagent validation covered failure/resume with repeated labels: the completed prefix retained its original child session, timestamps, and usage, while the next call created a distinct child session. Both persisted child headers retained the original parent after delivery-owner adoption.

PR #204, opened 2026-09-17
fix: report the real session model for agents that never resolve a spec

### LINKED ISSUE #202: Untagged subagents display the session's main model while actually running on the settings default
### What happens

An `agent()` call with no `model` and no `tier`, on a host without `~/.pi/workflows/model-tiers.json` (or without a `medium` entry), binds the pi settings default (`defaultProvider`/`defaultModel` in `~/.pi/agent/settings.json`) — by design since #36. But `/workflows`, the task panel, the persisted run record, and the journal all show the **orchestrating session's main model** for that agent, forever.

Observed on a real machine (v3.12.0): a session running on a non-default model ran a workflow of 8 untagged agents; the run JSON records the session's model for every agent while every call in fact went to the (different) settings-default model. Nothing anywhere in the run surfaces the substitution.

### Why

- `workflow.ts` seeds `displayModel = modelSpec ?? options.mainModel` and emits it on `onAgentStart`.
- The correction channel added in #167 (`onModelResolved` → `onAgentModel`) only fires inside `WorkflowAgent.run()` **when a spec was resolved**. For the untagged/no-config shape, `resolveAgentModelSpec()` returns `undefined`, no resolution happens, and the channel never fires — so the guess is never corrected, not even at `onAgentEnd` (it re-emits the same closure value).

#167's commit message already notes the residual gap ("on a fresh install, where onModelResolved does not fire"). This issue tracks exactly that gap: it is not merely cosmetic, because the persisted record misattributes possibly large fleets to the wrong model.

### Expected

Agents that never resolve a spec (untagged settings-default binding, the implicit-route degrade path, a requested tier that resolved to nothing) report the model the subagent session **actually bound**, through the existing `onModelResolved`/`onAgentModel` channel. Routing semantics unchanged — this is display/record honesty only.

### Proposed fix

After `createAgentSession()` succeeds, when `resolvedModel` is undefined, fire `onModelResolved` with `canonicalModelSpec(session.model)`. PR linked.

ArietidsZ:
Fixes #202.

Untagged agents (no model/tier, no medium tier configured) bind the settings default, and agents whose implicit route degrades land on the same fallback — but `onModelResolved` only fired for agents WITH a resolvable spec, so both shapes kept displaying the pre-resolution `mainModel` guess in `/workflows`, the persisted run record, and the journal for their whole lifetime (the residual gap #167's commit message calls out).

**Fix:** after the subagent session exists, when `resolvedModel` is undefined, report `canonicalModelSpec(session.model)` through `onModelResolved`. At most one fire per attempt/turn (the spec'd path sets `resolvedModel`, which guards the post-creation report), so the #167 invariants hold. The report carries the caller's thinking suffix (mirroring the spec'd path) and fires inside the session lifecycle so a throwing host callback cannot leak the session. Routing is unchanged.

**Tests:** end-to-end faux-provider test (settings default ≠ mainModel: asserts the run still binds the settings default AND the report carries it); thinking-suffix test; named-thread per-turn report test; throwing-callback disposal test; runWorkflow-level test locking onAgentStart guess → onAgentModel correction → onAgentEnd/onAgentJournal persistence; extended implicit-medium degrade test. `npm test` green (1461 tests, release gate 0 warnings).

Runtime-behavior verification per CONTRIBUTING: covered by faux-provider subagent sessions via the real `createAgentSession` SDK path; no real-provider environment run (flagging per the guide).


QuintinShaw:
Validated the updated head against current main. The change only corrects model telemetry after session creation; routing remains unchanged. The lifecycle try/finally also covers a throwing host callback.

Validation: 1,516 tests passed with Pi 0.86.1, and current-head CI passed. A real provider run used a fixture-only unavailable implicit tier to exercise the SDK settings-default path: the running model event, terminal agent record, and journal all reported the actual bound model and thinking level rather than the main-model display guess. Named-thread and callback-disposal regressions also passed.

PR #205, opened 2026-09-17
feat(routing): optionally inherit the main model for untagged agents

### LINKED ISSUE #203: feat: native option for untagged agents to inherit the session's main model (no model tiers required)
### Motivation

Hosts that launch pi with an explicit non-default model (CLI `--model`, `/model` switches, or agent hosts that pin a provider/model per thread) currently have no native way to make a workflow's untagged agents follow the conversation model:

- explicit `model` per `agent()` call — the script author (the LLM) must hardcode a model per task;
- `model-tiers.json` — works, but is a second, static configuration that drifts from the session and ranks by price hints, not by "what the user is talking to right now".

The session's current model is the natural default for untagged work, and it is what the UI already leads users to believe is happening (see the linked display-honesty issue).

### Proposal

A new optional user setting, default off (fully backward compatible):

```json
// ~/.pi/workflows/settings.json (or repo-local / project overlay)
{ "inheritMainModel": true }
```

When on, an untagged `agent()` (no `model`, no `tier`, no agentType/phase routing) inherits the orchestrating session's main model **as of run start** (a mid-run `/model` switch applies to subsequent runs), instead of the implicit `medium` tier / settings default. With no main model set, legacy routing applies. Explicit `model`/`tier` tags are unaffected. An inherited model that is unavailable degrades loudly via the existing `onModelFallback` path rather than throwing (same asymmetry as the implicit medium tier, since the script never asked for a specific model).

PR linked with implementation, tests, and README documentation.

ArietidsZ:
Closes #203.

New optional user setting `inheritMainModel` (boolean, default `false` = legacy routing), loadable from the global `~/.pi/workflows/settings.json`, the repo-local `<cwd>/.pi/workflows/settings.json`, or the external project overlay. When on, untagged `agent()` calls (no explicit model, tier, agentType model, or phase model) inherit the session's main model **at run start** (a mid-run `/model` switch applies to subsequent runs); with no main model set, legacy routing applies. An unavailable inherited model degrades to the settings default with a run-visible warning instead of throwing (same asymmetry as the implicit medium tier, since the script never asked for a specific model). Resume hash inputs are unchanged.

The `onModelFallback` payload gains a `source: "medium-tier" | "inherit-main"` discriminator so the run log names the route that actually degraded instead of always blaming the medium tier.

**Plumbing:** `workflow-settings.ts` (whitelist + normalize) → `pi-extension.ts` `buildManagerOptions` → `workflow-manager.ts` (constructor + `reconfigureAfterReload`, so settings reloads apply live) → `runWorkflow` options → `WorkflowAgent` → `resolveAgentModelSpec` (trailing options-object param `{ inheritMainModel?: boolean }`).

**Docs:** README settings paragraph (per CONTRIBUTING), capability-contract constraints, authoring-skill routing references (priority chain now places the inherited main model between metadata model and implicit medium); protected guidance re-accepted via `npm run guidance:accept`, generated files refreshed.

**Ordering note:** intended to land with or after the companion fix "report the real session model for agents that never resolve a spec" — without it, an inherit-degrade binds the settings default (logged) but the row display would keep showing the ghost main model; the capability constraint documents this either way.

**Tests:** settings normalize round-trip + invalid-value rejection + repo-local/project overlay precedence; manager field carriage + `reconfigureAfterReload` + an end-to-end manager-level run proving the flag actually routes the subagent (flag on → main model, off → settings default); `resolveAgentModelSpec` flag precedence (inherit beats implicit medium; explicit tags unaffected; no mainModel → legacy route); faux-provider runs proving the untagged agent executes on the main model with the setting on, that an unavailable inherited model degrades with `source: "inherit-main"` (and a no-main-model degrade reports `source: "medium-tier"`), and that the run log names the inherit-main vs medium-tier route correctly. `npm test` green (1468 tests, release gate 0 warnings).

Runtime-behavior verification per CONTRIBUTING: covered by faux-provider subagent sessions via the real `createAgentSession` SDK path; no real-provider environment run (flagging per the guide).


QuintinShaw:
Integrated the accepted runtime, scheduler, and RPC changes. The default remains unchanged; inheritance is opt-in and explicit routing still wins.

Validation: the integrated full suite passed 1,554 tests on Pi 0.86.1, followed by 164 routing/settings/policy checks after subsequent main updates. Final-head CI is green. Real-provider acceptance verified inherited main-model routing despite an unavailable fixture medium/default route, explicit-model precedence, and host-policy rejection before child-session creation.

PR #210, opened 2026-09-17
feat(middleware): opt child sessions into isolated trusted extensions

codesoda:
## Summary

Child workflow sessions intentionally load no host extensions (avoids per-child factory churn and recursive orchestration). However, some trusted extensions are wanted inside subagents too — e.g. provider adapters that must inject per-request headers, or hook-based helpers like `aislop` that give edit-time feedback on `edit`/`write` tool results. Without them, child sessions behave differently from the host session.

This adds an explicit opt-in allowlist, `providerMiddlewareExtensions: string[]`, for trusted middleware extensions in child sessions.

- **Default `[]`** — behaviour is unchanged unless the user opts in; no extension names are hardcoded.
- Configurable via `~/.pi/workflows/settings.json`, repo-local `.pi/workflows/settings.json`, project overrides, and SDK options (`WorkflowAgent`, `runWorkflow`, `WorkflowManager`).
- Names match resolved path components or an extension filename without its JS/TS suffix, case-insensitively; only enabled resources are considered, and project resources still require Pi project trust.
- **Recursion stays denied**: `pi-dynamic-workflows`, `workflow`, and `pi-subagents` are always excluded, even if allowlisted.
- Explicit SDK resource-loader injection remains authoritative and bypasses this discovery filter.

Example (`~/.pi/workflows/settings.json`):

```json
{ "providerMiddlewareExtensions": ["aislop"] }
```

## Motivation

Real-world case: the host session runs aislop, which reviews every `edit`/`write` tool result and feeds slop findings back to the model — but workflow subagents do most of the actual editing, and none of it gets that feedback because child sessions load no host extensions. Allowlisting `aislop` opts it into children, without opening child sessions up to arbitrary host extensions.

## Testing

- `npm test` passes on top of `v3.12.0` (unit tests, biome, docs check, context check, release gate — 0 warnings).
- New tests cover: exact-identity path matching (case-insensitive, no substring false positives), recursion denial even when allowlisted, package entrypoint resolution skipping disabled/recursive factories, and end-to-end child middleware binding (system-prompt + request-payload hooks observed by a faux provider) only when opted in.


QuintinShaw:
Fixed a child-session isolation problem found during validation: the SDK binds session actions into an extension runtime, so sharing an opted-in loader let one child append entries into another child. Opted-in factories/runtimes are now session-local, while extension-free loaders retain bounded sharing. Each created child also emits session_shutdown before disposal.

The allowlist now matches extension filenames or resolved package identities rather than arbitrary ancestor directories. Enabled-resource and project-trust checks remain in place; whole extensions must still be trusted, and explicit injected loaders remain host-owned.

Validation: a deterministic two-session regression failed before the fix and now passes; 1,601 tests and release gates passed on Pi 0.86.1, followed by 152 focused checks after UI integration. Two real-provider calls confirmed request hooks, session markers, and shutdown events stayed with their respective children. Final-head CI is green.

codesoda:
Nice fix, thanks for merging

PR #211, opened 2026-09-18
fix(rpc): show workflow lists and actionable background notices

codesoda:
## Problem

Running pi inside an RPC embedder (e.g. [Paseo](https://github.com/getpaseo/paseo)) exposes two host-mode bugs:

1. **`/workflows` silently does nothing.** The navigator is a `ui.custom()` TUI component, but the command gated it on `ctx.hasUI` — which is *also* `true` in RPC mode, where `custom()` returns `undefined` (pi's `docs/rpc.md` explicitly says to guard `custom()` with `ctx.mode === "tui"`, not `hasUI`). So in RPC hosts the command took the navigator branch, no-opped, and never reached the plain-text fallback sitting right below it.
2. **Start notices point at surfaces that don't exist.** `/code-review 10` (etc.) replies "watch the task panel or /workflows" — but the task panel is a `setWidget` component factory and the navigator is TUI-only, so in an RPC host the user is directed at two dead ends while their run is invisible until the final report.

## Changes

- `/workflows` (no args) and `/workflows ui` open the navigator only when `ctx.mode === "tui"`; every other mode prints the text run list via `pi.sendMessage`, which any host can render — verified working in Paseo, which displays `custom`-role messages as timeline entries.
- New `backgroundStartNotice()` helper (shared by built-in and saved workflow commands): TUI keeps the current wording; non-TUI hosts get pointed at `/workflows status <runId>` instead of the task panel.
- `/workflows watch` notice no longer promises "live progress in the status bar" outside the TUI.

## Tests

- RPC mode with `hasUI: true` prints the text list for both no-args and `ui` (regression for the silent no-op).
- TUI mode still opens the `ui.custom()` navigator and does not print the list.
- Mode-aware start-notice wording (TUI mentions the task panel; RPC mentions `/workflows status <id>` and must not mention the task panel).
- `npm test` green locally (1460 tests + release gate).

## Context

Found while running `/code-review` from Paseo. Follow-up work on the host side is tracked in [getpaseo/paseo#5031](https://github.com/getpaseo/paseo/issues/5031) (mapping `setStatus`/`setWidget`/`setTitle` to Paseo surfaces); a possible extension-side follow-up here is emitting string-line widgets in RPC mode (component factories are ignored by the RPC protocol), so hosts that render `setWidget` get the task panel too.

Marked draft for maintainer input on the non-TUI notice wording.


QuintinShaw:
Validated against Pi 0.86.1 in an actual isolated RPC process: extension commands registered, /workflows returned a plain-text list, and a saved workflow emitted an extension_ui_request notice pointing to /workflows status instead of an unavailable task panel. This surface test did not make a provider request.

The local full suite passed 1,546 tests, followed by integration with the accepted retry changes and focused command checks. Final-head CI is green.

codesoda:
Really excited for this one, can't wait to try it

PR #212, opened 2026-09-18
fix: seed persisted agents on resume to prevent ghost replay

### LINKED ISSUE #206: resume() wipes the persisted per-agent record to [] before replay rebuilds it
### What happens

`resume()` replaces the run's persisted per-agent record with `agents: []` on disk **before** replay rebuilds it (`workflow-manager.ts:1712` initializes the fresh `ManagedRun` with `agents: []`, `:1810` persists immediately, `writeRunToDisk` at `:1533-1544` serializes the empty array).

A paused run's on-disk record (per-agent model, tokens, sessionId/sessionFile, timestamps, errors) is destroyed at the instant resume begins. The in-memory maps built from `persistedAgents` keep the process-level resume working, but if the resumed execution dies before replay completes — script parse failure, early throw, crash — the durable loss is permanent.

### Evidence (reproduced)

Paused record with 1 agent (`model`, `status: "running"`) → `resume(id, { script: <unparseable> })` → immediately after `resume()` returns, `agents.length === 0`; after the run settles `failed`, the record has 0 agents and an empty journal — per-agent metadata unrecoverable.

### Fix direction

Carry the previous `agents` forward on the new `ManagedRun` (they are already captured in `persistedAgents` — seed `snapshot.agents` with them, marking them stale/replayed), or pass them through `writeRunToDisk` as a separate `priorAgents` section until replay has rebuilt. At minimum, don't persist until the first rebuilt row exists.

(Found during a full-repo audit of v3.12.0; reproduced with a faux-provider probe.)

ArietidsZ:
Fixes #206.

## Root cause

On resume, `runWorkflow` seeds in-memory agent state only from the journal's `agent_start` events. Agents whose start was persisted but whose completion landed after the last journal flush (or whose entry was written by a previous process) were missing from the seeded state, so the replay logic treated their persisted results as ghosts and re-executed them — burning tokens and, worse, silently double-applying side effects.

## Fix

Seed the replay state from **persisted agent records** (not just journal starts); replay dedupe resolves collisions by reverse-scanning for the **last** callId match, so a re-executed agent's fresh result supersedes the stale one instead of corrupting indices.

## Tests

- Ghost-replay token/timestamp preservation (mutation-killed J/K mutants)
- Per-field sanitization of persisted agent records; byCallId last-wins delete
- Unconditional ghost interrupt cause; clamp test via direct file write
- `VALID_PERSISTED_AGENT_STATUSES` satisfies-linked to the type
- Full suite 1465/1465; runtime change exercised e2e through real manager pause/resume tests

## Notes

- Backward compatible: old run files without the seeded records resume exactly as before.
- Known-not-fixed (documented): M2 row accumulation in a corner of the list view — cosmetic, deferred.


QuintinShaw:
Added a lease-protected full-record comparison so resume refuses a stale snapshot even when the persisted status is unchanged. A failed post-lease read now releases the lease. Regression coverage also verifies that a parse failure after resume preserves historical child-session metadata. The full local suite passed (1468 tests before integrating the shared-store fix); real Pi acceptance verified preserved session IDs/files across a failed edit and successful replay, with no duplicate dispatch or history rows. The updated head is undergoing CI.

PR #213, opened 2026-09-18
fix: persist auto-resume attempt counters

### LINKED ISSUE #207: Every manager persist erases autoResumeAttempts — usage-limit give-up cap resets on restart
### What happens

`PersistedRunState` carries `autoResumeAttempts` (`run-persistence.ts:171`), and the usage-limit scheduler writes it directly to persistence (`usage-limit-scheduler.ts:413`) and reads it on cold start (`:237,274`). But `writeRunToDisk` (`workflow-manager.ts:1484-1561`) builds its serialization literal **without** the field — so every manager persist (progress tick, pause, resume entry) erases it from disk.

Within one process the scheduler's in-memory state masks this; after an extension reload/session restart (exactly the span usage-limit waits cross), cold-start reads 0 → a fresh allowance of 5 auto-resume attempts with backoff. Repeated restarts mean the give-up sentinel can never stick: unbounded pause→resume cycles across sessions, each spending tokens.

### Evidence (reproduced)

Seeded file with `autoResumeAttempts: 2` round-trips through `load()`; one `resume()` → `persistRun` → field is `undefined` on disk.

### Fix direction

Add `autoResumeAttempts` to `ManagedRun` + the `writeRunToDisk` literal (scheduler updates it via the manager instead of a side-channel save); or merge-restore the field from the last read of the file. Add a round-trip test: every field in `PersistedRunState` survives a manager persist.

(Found during a full-repo audit of v3.12.0; reproduced with a persistence round-trip probe.)

ArietidsZ:
Fixes #207.

## Root cause

Auto-resume attempt counters lived only in memory. A process crash/restart (or a crossed resume path) reset the counter to 0, so a run whose provider kept failing could auto-resume forever — the configured attempt cap was unenforceable across restarts. Counters were also written unsanitized at four sites, so a corrupt persisted value (fractional, negative, NaN) could wedge the scheduler.

## Fix

- Persist `autoResumeAttempts` with the run state; restore on load.
- Single shared `sanitizeAutoResumeAttempts` (integer, >= 0) used at all four read/write sites.
- Human `resume()` resets the counter; auto-resume increments it — pinned by an e2e test through a real manager + scheduler.

## Tests

- Corrupt-counter sanitization cases; e2e human-resume counter-reset test
- Full suite 1464/1464

## Notes

- Backward compatible: runs persisted before this change resume with counter 0 (previous effective behavior).
- `fix/scheduler-robustness` is stacked on this branch's tip.


QuintinShaw:
Added a guard for paused or terminal ManagedRun entries that no longer hold a lease: counter updates now acquire a lease and merge only the counter into the fresh persisted record, rather than writing a stale snapshot. Regression tests cover a foreign owner. The original full suite passed (1465 tests); real Pi acceptance verified counter persistence across a controlled usage-limit pause, restart and automatic completion. The branch now includes current main and preserves both sets of appended regression tests; integrated CI is running.

PR #214, opened 2026-09-18
fix: shared-store discardDelta via per-key event-log undo

### LINKED ISSUE #208: SharedStore.discardDelta's Object.is guard deletes a sibling's equal-value write — live run diverges from resume replay
### What happens

`SharedStore.discardDelta` rolls back a failed agent's writes with an `Object.is` guard (`shared-store.ts:138`: `if (!Object.is(this.map.get(key), delta[key])) continue;`) — meant to leave a sibling's overwrite untouched. For equal **primitive/equal-value** writes the guard backfires: agents A and B both `store_put("k", "v1")`; B writes after A; A fails → rollback sees `Object.is("v1", "v1")` → the key is **deleted**, erasing B's live write.

B's journaled delta still contains `k: "v1"`, so resume replay restores it: siblings later in the *live* run see the key missing; the replayed run sees it — live state and replay state diverge.

Related: `store.get()` returns the **live reference** (`shared-store.ts:81-83`), so a read-modify-write re-put aliases identity across agents — the object-flavored version of the same hole.

### Evidence (reproduced)

The sequence above yields `store.has("k") === false` live, while `commitDelta(B) === { k: "v1" }` and `applyDelta` on a fresh store yields `k === "v1"`.

### Fix direction

Roll back from `priorValues` unconditionally for keys the window first-wrote, using the shadow chain (per-key last-writer ledger) instead of an `Object.is` current-value check; and/or make `store_get` return `structuredClone`d values. A store concurrency fuzz (random interleavings of two delta windows over shared keys with equal primitives) would have caught this immediately — existing tests only cover distinct-value interleavings.

(Found during a full-repo audit of v3.12.0; reproduced with a two-agent interleaving probe.)

ArietidsZ:
Fixes #208.

## Root cause

`SharedStore.discardDelta()` rebuilt prior state by replaying the whole put-history and dropping the tail — with interleaved writers (frames A/B alternating puts), discarding one frame's delta also silently dropped *other* frames' keys that happened to land in the discarded window, and the retention structure grew unboundedly within a window.

## Fix

Per-key write log ("event-log undo"): every put records (key, previousValue) so a discard walks backwards restoring exactly the keys that frame wrote — other frames' keys are untouched. A true per-window bound keeps at most the latest entry per window (moved to the top on overflow), so interleaved A/B writes no longer grow the log without limit. All mutation paths clone before mutate (trackPut/applyDelta/restore), keeping delta application atomic.

## Tests

- Interleaved A/B test: discard leaves length ≤ distinct-windows + 1 and reclaims at commit
- Fuzz oracle hardened (frozen pool + model-side clone)
- Two determinism tests contributed by an independent reviewer
- Full suite 1470/1470

## Notes

- Backward compatible: store semantics for non-discard paths are byte-identical.


QuintinShaw:
Validated the exact head with the full test suite (1470 passed) and a real Pi session on the latest SDK. The end-to-end case interleaves two same-value store writes, fails one writer after the other has been journaled, and verifies that a subsequent reader still sees the successful write. Independent review and CI passed. Merging this fix for #208.

PR #215, opened 2026-09-18
fix: estimated-usage flag follows the displayed figure

### LINKED ISSUE #209: commitWithFallback fabricates a token total with a zero breakdown and persists it as measured
### What happens

`commitWithFallback` (`agent-usage.ts:88-93`) fabricates a token total with a zero breakdown when a provider reports no usage (or an errored attempt): `total: ceil(len/4)` with `input/output/cacheRead/cacheWrite/cost` all 0. That estimate is persisted (`workflow-manager.ts:1547-1556`) as if measured — the record is internally inconsistent (`total ≠ sum of parts`), cost/budget displays understate real spend, and nothing marks the value as estimated (`tokenFigures` presents it as measured).

### Evidence (reproduced)

Tracker returns `{ input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 123, cost: 0 }`; end-to-end, an agent that never reports usage yields a persisted `tokenUsage.total === 2` with cost/input/output all 0 (estimate of `"hi"` + `"ok"`).

### Fix direction

Carry an `estimated: true` flag on fabricated usage; render estimates distinctly (e.g. `~2 (est.)`) and exclude them from cost-bearing totals; charge the prompt estimate only when the request plausibly reached the provider. A CI rule helps: any persisted total is either measured (`total === input + output + cacheRead + cacheWrite`) or explicitly flagged.

(Found during a full-repo audit of v3.12.0; reproduced with a faux-provider probe.)

ArietidsZ:
Fixes #209.

## Root cause

The `tokenTotalEstimated` flag was set on the *aggregate* when any contributor was estimated, but the **displayed** figure could come from a different composition (e.g. flagged agent sum vs unflagged small aggregate), so the UI could show a precise-looking number that was actually estimated — or vice versa.

## Fix

The flag follows the displayed figure: the display picks its source (unflagged small aggregate **or** flagged agent sum) and the estimate marker is computed from exactly that source. `tokenTotalEstimated` is optional on the wire; README documents the `~` convention.

## Tests

- Source-pick pinned (unflagged small aggregate + flagged agent sum)
- Pause-less commit-merge test; control-tool live-branch test
- Full suite 1469/1469

## Notes

- Backward compatible: absent flag renders as before.
- Documented limitation: the header-OR line is composition-masked (inputs + consumer pinned by tests); `agent.ts:484` streaming tag is untestable via the faux harness (faux computes usage); manager `commitFinalizedAgentUsage` and workflow `onTokenUsage` are two writers to `snapshot.tokenUsage` — both preserve the flag.


QuintinShaw:
Preserved backwards compatibility for formatter callers and added estimate provenance to the rolling token-rate samples. The rate remains marked when either endpoint is estimated, then clears the marker once both endpoints are measured.

Validation: 1,542 tests passed with Pi 0.86.1, current-head CI passed, and real provider checkpoint/restart acceptance retained the estimate flag in a mixed estimated/measured aggregate. The first response usage was deliberately withheld by the test adapter; this is not a claim that the provider omitted usage. A real Pi 0.86.1 TUI session also rendered the controlled detailed-panel fixture as ~300 tok and ~200 tok/s.

PR #216, opened 2026-09-18
fix: checkpoint suspension must not seal the run or drop paid results

ArietidsZ:
## Summary

A checkpoint suspension (`workflow.checkpoint()` awaiting human input) is an **intentional pause**, but the run-fatal seal path treated the drain that follows it as an abort: the run could be sealed as failed and already-paid-for agent results dropped.

## Fix

- Checkpoint suspension is exempt from both the run-fatal seal and the agent-completion `throwIfAborted` path.
- During the drain, a live `attachCheckpointResponse` attaches the human's reply; if the reply arrives after the in-memory window, it is persisted fail-closed with rollback.
- `pauseReason` precedence: `usageLimitPause` wins over `checkpoint` — a quota pause is never masked by a consuming/resumed checkpoint.

## Tests

- Pause/resume coverage uses in-flight agents whose first attempt rejects on abort; checkpoint tests release the sibling only after the drain's "waiting for N outstanding agent()" log appears
- Full suite 1459/1459

## Notes

- Backward compatible: no API change; suspension semantics unchanged for the happy path.
- Runtime change; e2e exercised through the manager pause/checkpoint test battery.


QuintinShaw:
Hardened checkpoint-response attachment so the live-record write requires its execution lease, and removed the redundant write after execution settlement, when that lease has already been released. Added regressions for lease ownership and the unleased fallback. The integrated local suite passed 1495 tests. A real Pi session confirmed that an attached response survives the sibling drain, the paid sibling result is journaled, and resume replays it without a second provider call.

PR #217, opened 2026-09-18
feat(scheduler): improve reset hints and bound resume retries

### LINKED ISSUE #207: Every manager persist erases autoResumeAttempts — usage-limit give-up cap resets on restart

ArietidsZ:
## Summary

Robustness cluster for the usage-limit scheduler (auto-resume of quota-paused runs):

- Absolute reset hints (`"Try again in ~299 min"`, pi-ai Codex wording) are anchored at **pause time**, not parse time — a hint parsed late no longer under-sleeps.
- `day`/`week` hint units parsed; jitter applied **before** clamping, with downward spread at the cap so a fleet of paused runs doesn't thundering-herd at the same instant.
- Refusal-poll backoff capped; cold start uses `max(prior, 1)` so a fresh process doesn't burn an attempt on a known-dead window.
- Equal-value writes are skipped (no journal churn).

## Tests

- Pin: no point mass at the cap (`every d < 3_600_000`)
- Wording extraction for "Try again in ~N min"; double-elapsed-subtraction BLOCKER found in review is fixed and pinned
- Full suite 1481/1481

## Notes

- **Stacked on #213** (`fix/persist-auto-resume-attempts`, fixes #207): this branch's base is that tip (`8acc9e9`), so the diff includes those commits until #213 merges. Only the scheduler commits belong to this PR; will rebase onto main after #213 lands.
- Accepted-as-noted: leftmost-match shadowing ("Try again in a moment. Resets in ~3h") is theoretical — noted, not fixed.
- Backward compatible defaults.


QuintinShaw:
Integrated the lease-aware attempt persistence changes and corrected cold-start delay accounting: compute the capped original delay before subtracting elapsed time, so restarts neither multiply elapsed time nor restart an already-capped wait. Malformed clock hints now fall back instead of being normalized into another reset time.

Validation: 1,571 tests and release gates passed against Pi 0.86.1; subsequent RPC integration passed focused scheduler checks, and final-head CI is green. A real-provider test used a controlled quota failure after a successful response, restarted the manager, and completed on the next provider call: the persisted armed-attempt counter remained 1 before restart, after restart, and at completion.

PR #218, opened 2026-09-18
fix: budget gate accounting across resume

ArietidsZ:
## Summary

Budget accounting fixes (audit cluster #1/#4/#5):

- **Budget gate moved after replay** (#1): the gate ran before journal replay, so a run that had exhausted its budget *before* the crash could never resume — the gate rejected it before replay could discover nothing billable remained. Gate now evaluates post-replay state.
- **Phase sub-budgets accumulate across resume** (#4): phase budget consumption is persisted (frame-namespaced `${frameRunId}:${title}` table); first declaration wins on conflict; nested-frame merge no longer overwrites.
- **Final usage flush moved after drain** (#5): the terminal usage snapshot was taken before the drain finished, losing the result payload's contribution; the flush now runs after all outstanding agents settle.

## Tests

- Exhausted-budget resume now succeeds; cross-resume phase-budget accumulation pinned
- Full suite 1465/1465

## Notes

- Backward compatible: runs persisted without the budget table resume with zeroed phase budgets (previous behavior).


QuintinShaw:
Validated the integrated head with 1500 passing tests and real Pi execution. A phase exceeded its soft token budget after its first real call; after restarting the manager, the completed call replayed without another provider request and the next paid call remained blocked. The original phase baseline and total usage were preserved. Independent call-path review and CI passed.

PR #219, opened 2026-09-18
feat(runtime): bound abort drain grace and finalize abandoned usage

ArietidsZ:
## Summary

`drainAbortGraceMs` (default 10s): after a run is aborted, the drain waits up to the grace for in-flight agents to settle before sealing — so a cooperative agent's result isn't dropped just because the user hit stop. A **non-aborted** drain still waits unboundedly (a checkpoint suspension is a deliberate pause, not a timeout candidate).

## Subtleties pinned by review

- The unbounded path must **not** race an already-resolved `abortWake` promise (that livelocked at 1.8M `allSettled` calls/3s, 102% CPU — found in r1, fixed and pinned).
- Option validation: `undefined` → default; `Infinity` → unbounded; finite `[1, 2^31-1]` honored (Node clamps larger values to ~1ms); anything else → default.
- A pause arriving **mid-drain** keeps the run paused with its result retained; interrupted agents are settled fail-closed before `persistRun`.
- The grace timer is intentionally ref'd (load-bearing for terminality); listeners removed in `finally`.

## Tests

- Full suite 1460/1460; livelock regression pinned

## Notes

- Accepted-as-noted (deliberate): `stop()` on a paused-with-result run keeps the result on the aborted record.
- Accepted-as-noted (roadmap): unattended no-abort wedge — a never-settling agent with no abort still parks the drain forever by design.
- README documents the new setting (CONTRIBUTING requirement).


QuintinShaw:
Added regression coverage and fixes for abort-drain abandonment: provisional usage is finalized before the last accounting flush, callbacks are closed before lease release, and queued calls cannot announce or start a new agent after abandonment. Checkpoint drains remain unbounded and preserve paid sibling results.

Real Pi 0.86.1 validation used one actual provider response deliberately held past a short abort grace. The run released its lease, retained the paid usage, ignored late callbacks, and suppressed the queued second call. This was a controlled delayed-response test, not a provider outage. The updated head includes current main and is undergoing final checks.

PR #220, opened 2026-09-18
feat(runtime): validate timeouts and add bounded retry pacing

ArietidsZ:
## Summary

Runtime polish cluster (audit #6–#9):

- **#6/#9**: mid-run `.md` deletion no longer wedges dependent agents (test uses a real mid-run deletion scenario); poisoned-result discrimination (`result.result === poisoned`).
- **Timeout validation split**: call-level `timeoutMs` (script-visible) throws synchronously for values outside finite `[1, 2^31-1]` (a `void agent()` call can't surface an unhandled rejection); run-level `agentTimeoutMs` (the persisted-resume surface) coerces invalid values — including legacy persisted `0` and `>2^31-1` — to the default with an `onLog` note, keeping legacy runs resumable.
- **`agentRetryBackoffMs`**: `0` disables backoff; negative/NaN/Infinity/throwing values fall back to default; finite values are clamped to a 2000ms cap (a larger cap parks the retry abort-unresponsive — up to 24.8 days at extreme values — and `setTimeout` overflows past `2^31-1` to ~1ms). Backoff stays abort-responsive.
- Lazy `commitWithFallback` thunk.

## Tests

- Sync-throw regression pinned via an `unhandledRejection` counter (mutation-verified)
- Full suite 1464/1464

## Notes

- `NavigatorState.reconcile()` is now single-param (the second param was unused); `noteManagerEvent` kept as a `@deprecated` no-op — `NavigatorState` is package-exported.
- Runtime change; e2e exercised through the runtime test battery.


QuintinShaw:
Integrated the current main branch and preserved estimated-usage provenance in the lazy fallback path. Closed or superseded attempts no longer evaluate fallback estimates; all-zero SDK usage retains the existing heuristic fallback behavior.

Validation: 1,552 tests and all release gates passed against Pi 0.86.1; the final-head CI passed. A real-provider acceptance run exercised a controlled recoverable failure followed by a successful retry, invalid-timeout rejection before spawning, and a nested agent definition remaining available after its source file was removed. No release is included.

PR #221, opened 2026-09-18
fix(ui): coalesce progress and preserve valid confirmations

ArietidsZ:
## Summary

UI hot-path efficiency + correctness cluster (audit #22–#32):

- **#22**: confirmations survive unrelated manager events — `confirm()` re-validates identity instead of wholesale cancel.
- **#23**: navigator event handler no longer eagerly calls `model.visible()`.
- **#24**: 100ms trailing-edge render coalescing; `flushProgress()` renders and covers every exit (including the `agentCount === 0` throw); single-pass Map phase bucketing; cached `Intl.NumberFormat`.
- **#25**: WeakMap `persistedToSnapshot` cache keyed by parsed-record object (rehydration no longer re-stringifies every agent result per accessor call).
- **#27/#28**: Set-seeded phase dedupe; lazy once-per-frame `selectedIdentity`.
- **#29**: incremental session-ACK scan (per-file offset cache, LRU-capped at 128, head fallback; `sessionFileContainsEntry` exported). The offset resumes at `cachedOffset - needle.length` — an off-by-one there made every repeat check degenerate to a full head scan (15.4ms → 0.03ms after the fix).
- **#31**: `maxAgents <= 0` falls back to 8; fractional floored (`slice(-0.5)` → `slice(0)` previously rendered ALL agents).
- **#32**: `hasActiveRun()` gate before the `settings()` disk read.

## Tests

- Full suite 1461/1461

## Notes

- **#26 (bounded preview serializer) was dropped per review**: 1.5–1.7× slower on wide objects with semantic divergences; `preview()` is byte-identical to main. The real per-frame cost is covered by #25.
- The t1 cost guard is probe-pinned (timing), not boolean-test-pinned.


QuintinShaw:
Added regression coverage for session-file scan boundaries, truncation, same-path replacement, and exact-record matching. Removed environment-specific timing claims from comments; this change does not claim a measured speedup.

Validation: 1,596 tests and release gates passed on Pi 0.86.1; after integrating the resource changes, 281 focused UI/tool tests passed and final-head CI is green. In a real Pi TUI, a controlled running task received progress events between the two stop-confirmation keystrokes and correctly transitioned to aborted. The UI fixture made no provider request.

Issue #37, opened 2026-06-26
feat: wire isolation field in agent frontmatter to createWorktree()


gr3enarr0w:
## Summary

`agent-registry.ts` parses the `isolation` frontmatter field but documents it as "parsed-but-ignored for now". Wire it to `createWorktree()` so users can declare `isolation: worktree` in `.pi/agents/*.md` without writing workflow scripts.

## Current state

```ts
// agent-registry.ts — documented as future work
// Parsed-but-ignored: mcp, skills, background, isolation
```

`createWorktree(baseCwd, name)` in `worktree.ts` is fully implemented.

## Proposed change

Add `isolation?: 'worktree' | 'none'` to `AgentDefinition`. In `WorkflowAgent.run()`, when the resolved agent type has `isolation: 'worktree'`, call `createWorktree()` before spawning and `removeWorktree()` in a finally block.

### Agent frontmatter example
```markdown
---
name: file-modifier
tools: read, edit, write, bash
isolation: worktree
---
```

## Files to change
- `src/agent-registry.ts` — add `isolation` to `AgentDefinition`, parse from frontmatter
- `src/agent.ts` — wire to `createWorktree()` in `WorkflowAgent.run()`
- `README.md` — document `isolation:` frontmatter field

Issue #36, opened 2026-06-26
Subagent model resolution falls back to first provider, ignoring configured default


QuintinShaw:
## Summary

When a workflow agent runs with **no explicit per-agent `model`**, the subagent model resolution does not honor the configured default in `~/.pi/agent/settings.json` (`defaultProvider: deepseek`, `defaultModel: deepseek-v4-pro`). Instead it falls back to the **first available provider** — in my environment `openai-codex/gpt-5.4`, which is expired and returns empty output (`AGENT_EMPTY_OUTPUT`).

Because every bundled workflow command (`/deep-research`, `/adversarial-review`, `/multi-perspective`, `/codebase-audit`) emits `agent(..., { label: '…' })` with no `model`, this means **all of them silently produce empty output** unless the session's resolved model happens to be a working one.

## Reproduction

Run any generated workflow through `runWorkflow` without setting a per-agent `model` (and without a model-tiers config). Observed via the agent-end `model` field:

```
start: performance [openai-codex/gpt-5.4]   ← NOT the deepseek default
ERROR: performance: Subagent produced no assistant output
```

Forcing an explicit per-agent model fixes it — all `deepseek/*` models (flash, pro, chat, v3.2) reply correctly when named:

```
start: performance [deepseek/deepseek-v4-flash]
end:   performance tokens=39538   ← real output
```

Passing `mainModel` to `runWorkflow` does **not** correct the fallback.

## Expected

A subagent with no explicit `model`/`tier` should resolve to the user's configured `defaultProvider`/`defaultModel` (here `deepseek/deepseek-v4-pro`), not the first available provider in the registry.

## Where to look

`src/agent.ts` — the `createAgentSession` call uses `SettingsManager.create(this.cwd, agentDir)` specifically to inherit `~/.pi/…/settings.json` defaults (per the comment around the empty-response fallback note), but the resolved model still comes out as the first available provider rather than the configured default. The default-resolution path (and `resolveAgentModel`'s untagged fallback) needs to actually read and apply the configured default.

## Impact

High — with default config, the bundled workflow commands appear to "run" (phases fire, agents start) but generate nothing. Easy to mistake for a broken model key when it's actually default-resolution picking the wrong provider.

## Notes

Found while real-pi-testing the `/multi-perspective` + `/codebase-audit` registration (#35). Orthogonal to that PR — those commands are correct; this is shared model-resolution behavior.

QuintinShaw:
[...]

PR #35, opened 2026-06-26
feat: register /multi-perspective and /codebase-audit commands

QuintinShaw:
Closes #33
Closes #34

## What

Both `generateMultiPerspectiveWorkflow` (`src/adversarial-review.ts`) and `generateCodebaseAuditWorkflow` (`src/deep-research.ts`) already existed and were exported, but neither was ever registered as a slash command — so the functionality was unreachable from the CLI. This wires both up as built-in commands, following the existing `/deep-research` and `/adversarial-review` pattern in `src/builtin-commands.ts`.

## Commands

- **`/multi-perspective "<topic>" [angle …]`** — analyze a topic from several independent angles in parallel, then synthesize. Defaults to `technical, product, security, user experience, maintainability` when fewer than two angles are supplied.
- **`/codebase-audit <scope> "<check>" …`** — run parallel checks over a scope, then cross-validate and report. Requires a scope plus at least one check.

## Notes

- Added a quote-aware `tokenizeArgs()` helper so a topic or check can span multiple words (e.g. `"missing error handling"`).
- `/multi-perspective`'s generator returns its prose under `synthesis` (not `report`), so the handler presents `synthesis` and falls back to the JSON dump otherwise.
- The two generator functions themselves are unchanged — this PR is purely the registration + tests + docs.

## Verification

- `npm run build` (tsc): clean
- `npm run check` (biome): clean
- `npm run test:unit`: **742/742 pass** (updated the count/idempotency/missing-only registration tests; added empty-arg validation tests for both new commands)

✅ Also verified end-to-end through the real SDK path on `deepseek/deepseek-v4-flash` — both commands ran with 0 errored agents and returned real non-empty `synthesis`/`report` prose. See the verification comment below.

PR #222, opened 2026-09-18
fix(host): close run watchers and preserve saved workflow backups

ArietidsZ:
## Summary

Host lifecycle cluster (audit #34–#37, #43):

- **#34**: `deleteRun` emits `"deleted"` when `deleted || managed` — in-memory runs whose file vanished out-of-band are included; `watchRun` subscribes.
- **#35**: a refused registration surfaces a warning instead of a false success.
- **#36**: `writeJsonAtomicPreservingPreviousBackup` keeps the *previous content* as `.bak` (version history semantics), but only when that content was actually read + parsed; an existing `.bak` is preserved only when it parses, else it mirrors the new content; an unreadable primary no longer fails the save.
- **#37**: try/catch around `setStatus` + `sendMessage` on teardown (non-vacuous test: status "running", `setStatus` throws only on teardown).
- **#43**: shadow-builtin argument contracts — `SHADOW_WHOLE_STRING_PRIMARY` (deep-research→question, adversarial-review→task) from `_raw` (exactly `args.trim()`, so `"="-containing topics survive); `SHADOW_TOKENIZED_PRIMARY` (multi-perspective→{topic,perspectives}, codebase-audit→{scope,checks}) via quote-aware `tokenizeArgs`. code-review is unmappable (its builtin diff comes from git capture) and stays unshadowed. Positional args beat parameter defaults; explicit `key=` beats positional; a leading `k=v` token skips tokenized mapping.

## Tests

- Full suite 1467/1467; manager-level pin for the deleteRun emit (added after a mutation-restore accident destroyed the uncommitted change)

## Notes

- Runtime change; e2e exercised through the command/host test battery.


QuintinShaw:
Verified against the current main branch and added regression coverage for corrupt primary/backup combinations and backup-write failure. Deletion still respects the existing lease guard and releases its lease even when a listener throws; completed watchers unsubscribe before touching a stale UI context.

Validation: 1,624 tests passed locally with Pi 0.86.1, including context and release checks. Real Pi RPC validation confirmed saved-builtin positional arguments, status responses, the reserved-name registration warning, and recovery of the previous saved version. The RPC command-surface test made no provider requests. Required CI passed on 0239811d0a701ebc9127b46681ca79591b02c9ad.

Issue #41, opened 2026-06-26
feat: add confirmation dialogs before destructive actions in navigator


gr3enarr0w:
## Summary

Pressing `x` in the navigator immediately stops a running workflow or deletes a saved workflow with only a post-hoc `ui.notify()`. A confirmation step prevents accidental data loss.

## Current state

```ts
case 'stop':
  manager.stop(action.runId)     // immediate, no confirm
  ui.notify(`Stopped ${action.runId}`)
  break
case 'deleteSaved':
  storage.delete(action.savedName)  // immediate, no confirm
```

`ctx.ui.confirm()` already exists and is used in `/workflows-models` — infrastructure is in place.

## Proposed change

```ts
case 'stop': {
  const ok = await ctx.ui.confirm(`Stop workflow "${run.name}"? This cannot be undone.`)
  if (!ok) break
  manager.stop(action.runId)
  break
}
case 'deleteSaved': {
  const ok = await ctx.ui.confirm(`Delete saved workflow "/${action.savedName}"?`)
  if (!ok) break
  storage.delete(action.savedName)
  pi.unregisterCommand(action.savedName)
  break
}
```

Also apply to `pause` with a lighter confirmation.

## Files to change
- `src/workflow-ui.ts` — wrap `stop`, `deleteSaved`, `pause` with `ctx.ui.confirm()`

QuintinShaw:
Thanks — #164 adds identity-bound double-key confirmation for pause, stop, and delete, so list refreshes cannot redirect a pending destructive action. It will close this issue when merged.

PR #223, opened 2026-09-18
feat(resources): bound child loaders, web bodies, and git operations

ArietidsZ:
## Summary

Agent-layer reliability cluster (audit #21/#41/#42):

- **#21**: all four `execFile("git", …)` calls now run with `{timeout: 30s, maxBuffer: 4MB}` plus an injectable `WorktreeExecOptions.timeoutMs` (0 disables). A timed-out/failed `worktree add` no longer leaks: git rolls back its registration but never deletes the `-b` branch or the partial tree — cleanup now deregisters first (`worktree remove --force`), then removes the tree with retries (a SIGTERM'd git keeps writing for a few hundred ms; unretried `rm` aborts on the first ENOTEMPTY), prunes stale registrations, and only then deletes the branch. Failure reasons are honest: a killed git reports "git timed out (slow or hung filesystem?)", not "not a git repository".
- **#41**: the shared `DefaultResourceLoader` memo is LRU-bounded at 8 entries (touch-on-hit keeps the hot base-cwd entry resident), pruned at turn end; the failure path clears the memo identity-checked so a stale rejection can't evict a newer healthy build. Worktree isolation previously retained N fully-reloaded loaders until run end.
- **#42**: `web_fetch` streams the response body with a 2MB cap (`readBodyCapped`, exported) instead of buffering the whole body and running ~10 regex passes on the event loop. Charset/BOM/204 parity with `res.text()` preserved; the cap is on decoded bytes.

## Tests

- Hung-git PATH-shim test (150ms bound); timed-out-add cleanup test with **real** residue (shim runs the real add, then hangs) — mutation-verified, deterministic under load (timeoutMs 2s), asserts branch removal, no leftover tree entries, no stale registration
- 5MB-body cap test; loader-memo bound test
- Full suite 1460/1460 across repeated runs

## Notes

- Accepted platform limit: SIGTERM-only kill cannot bound a D-state/uninterruptible child (network FS); the credential-helper class of hangs is bounded by the timeout.
- Cleanup worst case is bounded (~15s of best-effort 5s-timeout steps) and failure-tolerant.
- `WorktreeExecOptions` is now exported from the package root.


QuintinShaw:
Tightened two lifecycle boundaries during review: failed worktree creation now removes only its generated path/branch and never prunes unrelated registrations; the loader memo is capped immediately on insertion, including pending builds, with identity-checked failure eviction.

Validation: 1,592 tests and all release gates passed on Pi 0.86.1, and final-head CI is green. A real-provider workflow ran inside a fresh Git worktree and removed its temporary worktree/branch afterward. A real local HTTP response larger than the cap was limited to 2 MiB and released its reader. The timeout regression uses real Git residue and verifies that an unrelated missing worktree registration survives cleanup.

Issue #39, opened 2026-06-26
feat: add search/filter to the /workflows navigator


gr3enarr0w:
## Summary

The `/workflows` navigator lists all runs and saved workflows with no way to filter. With many runs this becomes unwieldy.

## Proposed design

- Press `/` on the runs view to enter filter mode — a text input appears at the top
- Typing filters runs and saved workflows by name, runId, and status (case-insensitive substring)
- `esc` exits filter mode; `esc` again with empty filter clears it
- Filter indicator shown in header: `─ Workflows · filter: "deep-research" ─`

### `NavigatorState` change

Add `filter?: string` to `StackFrame`. Filter is applied before rendering, cleared when drilling into a run.

## Files to change
- `src/workflow-ui.ts` — `NavigatorState`, `renderNavigator`, `keyToAction`, `openWorkflowNavigator`

QuintinShaw:
Thanks — #164 adds live filtering by run name, run ID, status, and saved-workflow metadata while keeping the selected identity stable across live refreshes. It will close this issue when merged.

Issue #33, opened 2026-06-26
feat: register /multi-perspective as a built-in slash command


gr3enarr0w:
## Summary

`generateMultiPerspectiveWorkflow(topic, perspectives)` already exists in `src/adversarial-review.ts` and is exported from `src/index.ts`, but it is never registered as a slash command. This is a two-line fix in `builtin-commands.ts` that exposes a genuinely useful built-in.

## Current state

```ts
// src/adversarial-review.ts — exists, exported, but unreachable from the CLI
export function generateMultiPerspectiveWorkflow(topic: string, perspectives: string[]): string
```

`src/builtin-commands.ts` only registers `/deep-research` and `/adversarial-review`. `/multi-perspective` is invisible to users.

### `src/builtin-commands.ts`

```ts
pi.registerCommand('multi-perspective', {
  description: 'Analyze a topic from multiple independent perspectives in parallel, then synthesize',
  usage: '/multi-perspective <topic> [perspective1] [perspective2] ...',
  handler: async (args, ctx) => {
    const [topic, ...perspectives] = parseArgs(args)
    const defaultPerspectives = ['technical', 'product', 'security', 'user experience', 'maintainability']
    const script = generateMultiPerspectiveWorkflow(
      topic,
      perspectives.length >= 2 ? perspectives : defaultPerspectives
    )
    const result = await runWorkflow(script, {}, { cwd: ctx.cwd, tools: createCodingTools(ctx) })
    ctx.sendMessage(result.report ?? JSON.stringify(result))
  }
})
```

### Default perspectives

When the user doesn't supply perspectives, default to `['technical', 'product', 'security', 'user experience', 'maintainability']` — useful for any code/design decision.

### Usage examples

```
/multi-perspective "should we use Redis or Postgres for session storage"
/multi-perspective "JWT vs session cookies" security scalability developer-experience
/multi-perspective "refactor auth middleware" technical maintainability risk
```

## Files to change

- `src/builtin-commands.ts` — register the command
- `README.md` — document `/multi-perspective` with examples

## Notes

The workflow generator already handles the parallel fan-out and synthesis. This issue is purely about discoverability — making the function reachable from the slash command interface that users actually see.

QuintinShaw:
[...]

PR #224, opened 2026-09-18
fix: persistence & lifecycle reliability (deleteRun lease gate, delivery, logger)

ArietidsZ:
## Summary

Persistence & lifecycle reliability cluster (audit #16/#19/#33/#38/#39 + two deeper holes found in review):

- **#16 — cross-process `deleteRun` resurrection**: deleting a run whose lease a live foreign process holds used to succeed; that process's next persist silently resurrected the record. `deleteRun` now acquires the run lease first and refuses while a live process holds it. Review found the managed branch still bypassed the gate: a paused/terminal in-memory run released its lease at pause settle, so a foreign **resume** could hold the lease while an in-memory `deleteRun` sailed through (two-process repro). The gate now applies whenever the entry doesn't actually own its lease; the lease is held *across* the delete (closing the release→delete TOCTOU); probe failures refuse gracefully (no-throw contract restored); `/workflows rm` distinguishes "active in another live session" from "No run".
- **#19 — pause-settle grace 1s → 10s**: a human's checkpoint reply failed with "still settling" when the pause tail (a full-state persist on a slow/synced disk) exceeded 1s. Injectable via `_setPausedExecutionSettleTimeoutForTests`.
- **#33 — result-delivery handler leak/regression**: `turn_end` registration was per-manager with no `off()`, so every cross-project `session_start` rebuild stacked a handler closing over the stale manager (rooting it for the process lifetime). Registration is now once per **ExtensionAPI instance**, dispatching to the current manager — the guard is keyed to the pi instance because the cached dist module + per-generation fresh pi would otherwise leave generations 2+ with no handler at all (found in review).
- **#38 — lazy fallback manager**: `createWorkflowTool()` eagerly constructed a `WorkflowManager` whose constructor scans — and on stale-run recovery, **rewrites** — the cwd's run store. Schema-only callers (context measurement, release gate) no longer touch the developer's real global store.
- **#39 — logger write amplification + silent-loss edge**: `persist()` rewrote the whole log file per call (2× write volume per pause/resume cycle; a resumed run's fresh logger also truncated the earlier execution's lines). Now append-only with **per-entry persisted flags** — a watermark alternative would silently drop lines whose write-through append failed transiently before a later success (found in review).

## Tests

- Lease-gate tests (unmanaged + terminal-in-memory + foreign lease), no-throw contract, rm message discrimination, per-generation turn_end, logger transient-failure recovery — mutation/dual-version verified (0/5 pass on the pre-fix commit, 5/5 here)
- Full suite 1464/1464

## Notes

- Accepted-as-noted: a truncated half-line from an interrupted ENOSPC append can merge with the next appended line; `persist()` doesn't repair the seam (diagnostic-only).
- Backward compatible: no schema change; legacy run files and log files behave as before.
- Runtime change; e2e exercised through the manager/commands/delivery test batteries.


QuintinShaw:
Added focused fixes during review: keep all primary/legacy recovery records protected until deletion is complete; register turn_end per Pi instance and refresh a same-session manager binding without replacing its transport; append logger entries in order across transient write failures, preserving earlier resume history.

Validation: 1,522 tests passed against Pi 0.86.1 before integrating the separately validated model-reporting change. Real provider acceptance confirmed checkpoint/replay, preservation of the prior log prefix after restart, refusal to delete a foreign-leased run, and successful terminal deletion without a recoverable run record. Independent review found no remaining blocking issue in these paths. Current-head CI is running.

